PR #15, opened 2021-10-03
HxButton.Color defaults

### LINKED ISSUE #14: Not specifying HxButton.Color breaks HxButton
HxButton throws an exception when the Color attribute is not specified. One should assume if the color is not specified, it's display properties will be controlled in the CssClass property.

robertmclaws:
Fixes #14


hakenr:
If we decide to set default for `HxButton.Color`, the setting should go to `ButtonDefaults` class as initial value for `Color` property (will get to HxButton as default value for `HxButton.Defaults.Color`).

robertmclaws:
That's fine, I set it there as well, so I can remove the redundant assignment. Would you like me to submit another change to the PR?

hakenr:
We have a library-design-meeting on Monday where we will discuss this issue. I will adjust you PR as needed then or let you know if we decide to stick with the current state. In the meantime, you can set the `HxButton.Defaults.Color = ThemeColor.None` in your `Program.cs` to set your own default color for the whole project.

Issue #17, opened 2021-10-03
[HxButton] Cannot specify HxTooltip span CSS classes


robertmclaws:
I have a button group I'm implementing where I need to be able to set the padding on the button content. However, when a Tooltip is specified, the HxButton.CssClass is not set on the `<span />` used to contain the tooltip, so it renders incorrectly.

![image](https://user-images.githubusercontent.com/1657085/135740793-2fc631d4-2e2b-47e1-a54e-cb4fa78b8f73.png)
The one on the top is the proper rendering, using the original controls I'm replacing. the one on the bottom is with HxButtons. The first button does not have a tooltip specified, whereas the other 2 do. So `px-2` is set on the `<button />`. not the `<span />` that contains it.

hakenr:
Hi @robertmclaws, we are aware of this issue and investigating the possibilities to fix.
Current implementation of HxTooltip (also used in HxButton) does not render the span at all if the `HxTooltip.Text` is not set.
The was introduced to make the HxButton as lightweight as possible (pure `button` element) if there is no `HxButton.Tooltip` set (which we believe is the by far most common case).
The current implementation has to be changed, but still not sure in which direction:
1. drop `HxButton.Tooltip` support
2. always render the HxTooltip's `span` (and set the `HxButton.CssClass` to it)

Can you help us with arguments for one or another direction? Or any third one?

robertmclaws:
I think the current implementation is fine, TBH. I liked the thoughtfulness of how it worked. I believe my PR handled it efficiently in a way that allows you to specify CSS classes on both the `<button>` and the `<span>` predictably.

I can see an argument for always rendering the span, though that might bother people trying to achieve 1:1 HTML ports.

I think the documentation just needs to be clearer inside the Button Tooltip demo that adding a tooltip will wrap the button in a span.

hakenr:
We have to fix this together with the implementation of HxTooltip. The current behavior is weird, if you do not set `HxTooltip.Text` nothing gets rendered except the child content. What if you set `HxTooltip.SpanCssClass` and leave the `HxTooltip.Text` empty? Do you expect the `SpanCssClass` not being used at all?

robertmclaws:
It depends on whether or not you render the span all the time. Not rendering the container if the Text is empty makes a lot of sense, so I would keep that behavior.

[...]

PR #18, opened 2021-10-03
[HxTooltip] CssClass & SpanCssClass

### LINKED ISSUE #16: [HxTooltip] CssClass property missing
There is no way to control the rendering of a Tooltip, for example to change the mode from Light to Dark. In the documentation, this is controlled by the `data-bs-custom-class` attribute, but the HxTooltip control does not implement attribute splatting, so you cannot set this manually.

robertmclaws:
Fixes #16, and allows for #17 to get resolved properly in a separate PR.

- Creates a CssClass property, which the end user would assume applies to the Tooltip itself.
- Creates a SpanCssClass property, so the developer clearly sees that class is applied to the container.
- If a span is rendered, combines the current "d-inline-block" class with the `SpanCssClass` property.
- Utilizes the Bootstrap documented `data-bs` behavior to add CssClass values to the tooltip rendering.


hakenr:
In the end I decided to go for `WrapperCssCass` as the term "container" is already used in [Bootstrap tooltip API](https://getbootstrap.com/docs/5.1/components/tooltips/#options) (might be implemented later in HAVIT Blazor) and reusing the same term might introduce some confusion for those familiar with the original usage of the "container".

robertmclaws:
Thanks for merging this!

PR #44, opened 2021-11-21
Add mobile keyboard support for decimal inputs

robertmclaws:
If there is 1 or more decimal place, then the keyboard opened on mobile should have the decimal key present.


hakenr:
Hi @robertmclaws,
my colleague @jirikanda might add more details, but I think it is a intentional decision not to use mobile decimal keyboard for decimal inputs as it does not allow you to enter decimal numbers when the globalization (CultureInfo) of the application is not the same as keyboard chosen by the user (e.g. in Czech Republic the numeric format is 1 000,00 instead of 1,000.00).

But we might consider making this configurable...

hakenr:
@robertmclaws, we made the InputMode configurable. Will be part of next release.

PR #52, opened 2021-12-19
Remove d-inline-block from `HxTooltipInternalBase.cs`

robertmclaws:
This class interferes with developers wanting to use flex layouts with buttons.


robertmclaws:
Hey folks! I' trying to build button group layouts, and this class is jacking with my code. If you folks need that class on your wrappers, then you should set it yourselves with the TooltipWrapperCssClass property, IMO.

hakenr:
Hi @robertmclaws,
I fully understand your frustration. The original implementation originated with this issue:

In `HxButton` there is the `Tooltip` feature where you are able to set the tooltip for the button. This uses the `HxTooltip` wrapper where there is a conditional rendering of the wrapping `span` which gets rendered only if the tooltip is set. This causes weird bahavior as the depending whether the `HxButton.Tooltip` is set, the button behaves as `inline-block` (when Tooltip not set) and as `inline` (when Tooltip is set and the wrapping `span` gets rendered).

We already met this issue in our applications and were able to resolve it with additional explicit `d-xy` class set to `HxTooltip.CssClass` which will override the existing `d-inline-block`. 

Now, from the broader view, it seems to me that the `d-inline-block` should be set to the `HxTooltip` from the `HxButton` to isolate it only for buttons where we want to achieve the unification of the basic display behavior with buttons without tooltip (to prevent changing the layout depending on tooltip being set or not). But still, it should be there for the buttons...

Or do you thing we should drop this requirement and let the users set `HxButton.TooltipWrapperCssClass` on their own when using the tooltips? From the very first feedback we had from our internal testing, this seemed to be quite confusing for the developers and they spent quite a noticeable time on investigating the introduced layout issues (as it did not come to their might that it can depend on the `HxButton.Tooltip` being set - which can even be dynamic).

cc @crdo

robertmclaws:
> From the very first feedback we had from our internal testing, this seemed to be quite confusing for the developers and they spent quite a noticeable time on investigating the introduced layout issues (as it did not come to their might that it can depend on the HxButton.Tooltip being set - which can even be dynamic).

I think the challenge is that HxButton sets the value, the _exact opposite_ situation is true: I spent quite a bit of time trying to figure out where that class was coming from. For example, if you are trying to `d-flex justify-content-center` the container, and then `flex-fill` the buttons so they have equal width, that class throws things off without also specifying a `d-flex` to override the HxButton default.

Maybe the solution here is that if nothing is specified, set `d-inline-block`. If the user provides a `TooltipWrapperCssClass` value, then they are taking responsibility for the display, and potential need for that class should be noted in the docs and the `<remarks>` code comments for Intellisense.

hakenr:
I would like to make it as smooth as possible and setting the class if there is no other class set does not make sense for me as the usual value for `TooltipWrapperCssClass` is something like `mb-2` and such. Changing the layout by setting the margin would be confusing as well.

I'm just wondering, the `.btn` class has `display: inline-block;` in its definition. Is it the `!important` modifier which makes the difference in the `d-inline-block`? Is there any other `.bnt`-something selector which covers some specific cases? Can you please provide more details on your scenario so we can think of the best solution?

hakenr:
Hi @robertmclaws,
we had an internal discussion and came to a state where we are not sure what exact issue we are fighting with (except the fact that the `.d-inline-block` should be used only for tooltips wrapping the `HxButton` and possibly should not use the `!important` utility class but a some other less aggressive one).

We think the button-group issues you are describing are not related to the `inline-block`, but the plain fact there is the wrapping `span` which changes the structure and breaks the `HxButtonGroup` when used with tooltips (the Bootstrap selector there requires first child should be `.btn` and we will try to get over this with some additional CSS rules with released selectivity).

The same applies for the flex-related issues where you want the buttons to have same width. It is not the `inline-block` but the structure itselft with `span` which makes things complicated. But you should still be able to achieve what you need (e.g. by setting `HxButton.CssClass="d-block"` etc.).

Can you please provide us with repro-steps to be sure we are getting your issue?

cc @crdo, @jirikanda

PR #90, opened 2022-03-09
Add HxMessageBox.zh-CN.resx

linjianlin:


hakenr:
@linjianlin Thanks for the contribution. Is there any chance you will add more translations? There are not many .resx files to translate, I guess there are less then 100 items altogether to translate.

PR #128, opened 2022-05-02
Fix popover on safari

victornor:
Using data-bs-container breaks support on safari.
Furthermore, trigger hover only works when used in conjunction with focus.


hakenr:
@crdo, can you please review?

hakenr:
@crdo, we should probably still make the container configurable to allow handling scenarios such as overflow in responsive tables?

victornor:
Hey, i have worked more on this and it turns out, that the issue was caused by using Trigger="PopoverTrigger.Focus".
This breaks both IPhone and Mac (on safari) support for some reason.
Omitting this or using PopoverTrigger.Hover seems to have solved the issue.

hakenr:
cc @crdo

PR #138, opened 2022-05-06
JSRuntime.ImportModule() extension method

hakenr:
@jirikanda Take a look at `GetAssemblyVersionIdentifierForUri()` please. Is it the approach we want to use to get the right version info?


hakenr:
(Merged, but the review is still appreciated.)

Issue #117, opened 2022-04-15
[HxDropdown] vs. .btn-group => [HxDropdownButtonGroup] component?


hakenr:
Current implementation of `HxDropdown` emits `.btn-group` CSS class except a few cases (e.g. inside `HxNavbar`).

This was a good decision as far as the dominant scenario was creating buttons with dropdown menus.

This turned out to be an issue for scenarios with custom `HxDropdownToggleElement` where much more complex component is to be assembled (such as `HxSearchBox`). The `btn-group` class is not wanted in such scenarios and causes unwanted effects:
![image](https://user-images.githubusercontent.com/12828888/163650786-2bc818da-31cf-4e7f-8e39-9217b83bd1bb.png)
(the width is restricted etc.)

As we want to support using `HxDropdown` for such complex scenarios, we have to allow omitting the `btn-group` class.

## Failed attempts
I tried to remove the `btn-group` from `HxDropdown`. This literary breaks all the dropdown-buttons where the `btn-group` has to be added manually (e.g. as `CssClass="btn-group"`). This is not intuitive and would cause many headaches to the users.
1. Adding a `IsButtonGroup` parameter to the `HxDropdown` is weird. We could set the default value to `true` but it is far from being ideal.
2. Trying to detect `HxDropdownToggleButton` in child-componets is tricky and would lead to in-natural component lifecycle. It is justifiable for components such as `HxGrid`+`HxGridColumn`, but I don't like the idea for such "raw" component as dropdowns.

## Proposed solution
I propose to distinguish the two scenarios in separate components. 1. plain `HxDropdown` (without `btn-group`) and 2. `HxDropdownButtonGroup` for buttons (damm, `HxDropdownButton` sound much more natural but it is not what it is).
* the `Split` parameter of current `HxDropdown` will be also moved to `HxDropdownButtonGroup` as it is also button-specific,
* it is a breaking change, but it will be anyway.
* I'm not sure whether to require `HxDropdownToggleButton` to be placed inside the new `HxDropdownButtonGroup` or whether to allow plain `HxDropdown`. The first will make the migration of the breaking change more explicit as it will require all the usages to be adjusted to run (in opposite to just breaking the visual experience).

@crdo @jirikanda @robertmclaws What do you think of the above? Any other suggestions?

crdo:
@hakenr Sounds too complicated to me. You only need the `.btn-group` class if the `Split` parameter is `true`. The regular button does not need it at all. Or am I missing something?

PR #149, opened 2022-05-13
#117 [HxDropdown] vs. .btn-group => [HxDropdownButtonGroup] component?

hakenr:
Please review.

Should we set `role="button"` to `HxDropdownToggleElement` by default?


hakenr:
Nobody cares, I'm merging then... :-P

Issue #78, opened 2022-02-15
[HxSidebar] Expand section if current location matches HxSidebarItem href


crdo:
If I directly navigate to some page that matches some HxSidebarItem, I would like to have its section open (expanded).

hakenr:
@crdo Should be achievable with pure CSS?

hakenr:
@Harvey1214 wanna give a try? (consult the intended approach with @crdo)

Harvey1214:
@hakenr ofc

hakenr:
Current implementation instructions from @crdo:
1. remove `aria-expanded="false"` in `HxSidebarItem#39` if the item is active https://github.com/havit/Havit.Blazor/blob/da0c36698ff63aa2e0a3ce4215fd270916c595f9/Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor#L32-L39
2. remove `collapsed` class in HxSidebarItem#64 if the item is active
https://github.com/havit/Havit.Blazor/blob/da0c36698ff63aa2e0a3ce4215fd270916c595f9/Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor#L62-L69

Will be a little tricky as we will have replicate the active-detection logic in HxSidebarItem (encapsulated in `NavLink` as private methods).

crdo:
@hakenr sorry, 2. should be to add class `show` instead of removing `collapse`.

PR #150, opened 2022-05-13
[HxSidebar] Expand section if current location matches HxSidebarItem href #78

Harvey1214:


hakenr:
@crdo Please take a look if this is what we wanted. I was hoping for a CSS-only solution.

hakenr:
@Harvey1214 We agreed on Blazor & CSS solution with @crdo, no need for JS. I have some instructions from @crdo and will give it a try.

Issue #154, opened 2022-05-17
[HxMultiSelect] Hover cursor shape should be default instead of text


hakenr:
![image](https://user-images.githubusercontent.com/12828888/168745343-56d0b26a-2c22-42d3-afb8-21bb0654ad1f.png)
(wrong proportions in image)

Harvey1214:
PR #155

PR #155, opened 2022-05-17
[HxMultiSelect] Set cursor on hover to default instead of text #154

Harvey1214:
@hakenr Do you think this is a viable solution? It achieves pointer cursor on hover but also centers the text, however that can be easily adjusted if need be.


hakenr (commented):
@crdo Can you please review? Not sure the `<input type="button">` is an ideal approach for cursor shape here.

crdo:
From the accessibility POV not. I'd go for rule in css isolation.

hakenr (changes requested):
@Harvey1214 Please revert and set the cursor shape with appropriate CSS rule (CSS isolation).

Harvey1214:
@hakenr completed

PR #157, opened 2022-05-19
Bootstrap icons update

Harvey1214:


Harvey1214:
@crdo You're correct, however, there were no changes made to the JSON file you are referring to, therefore, there is no need to update it.

PR #159, opened 2022-05-19
#78 [HxSidebar] Expand section if current location matches

hakenr:
I have created a very first prototype.

Current findings (work-in-progress):
1. Although I comment-out calling `StateHasChanged()` from `OnAfterRender`, the `HxSidebarItem` still re-renders (3x render!!!) and correct output gets rendered.
2. The server-prerendered output reflects the expansion, which means at least 2x rendering has had to occur on server-side.
3. We need more granular setting. E.g. for our documentation, the `HxBadge` component is present three times in the Sidebar (Buttons&Indicators, Layout&Typography, All components), which means all three sections get expanded.

cc @crdo; @jirikanda


hakenr:
@crdo Give it a try, if you have a minute. I will adjust the doc sidebar + merge later afternoon and create a new release.

PR #160, opened 2022-05-19
HxMultiSelect - ToggleTemplate?

Harvey1214:
@crdo I'm suggesting adding the possibility to define custom content for the element that is used for opening the HxMultiSelect. An example use-case would be if we'd want to let the user show/hide grid columns using HxMultiSelect and show only a button with an icon to open the dropdown menu.

If we want to render a RenderFragment as the content, then we are going to have to use a different element than "input".

Thanks for considering.


hakenr:
I currently don't see much value in replacing the whole input (toggle) with something completely new. Closing for now, but will keep the branch for "maybe in future"...

Harvey1214:
Sure, I think it would be beneficial though to be able to set at least a custom text instead of displaying all the selected values. I find it's quite a common scenario to be able to choose which columns of a grid will be visible and which hidden and multiselects are often used for that. For example in Shopify.

image

That's why I figured we might want something like that for our projects working with large amounts of data.

hakenr:
Makes sense (text customizability). Please create a separate issue and we will "design" the feature there.

PR #163, opened 2022-05-20
ValidationMessageMode

hakenr:


jirikanda:
@hakenr I would like to ask for review for the "effective" cascade:
```
protected ValidationMessageDisplayMode ValidationMessageDisplayModeEffective => this.ValidationMessageDisplayMode ?? this.GetSettings()?.ValidationMessageDisplayMode ?? GetDefaults().ValidationMessageDisplayMode ?? HxInputBase.Defaults?.ValidationMessageDisplayMode ?? throw new InvalidOperationException(nameof(ValidationMessageDisplayMode) + " default for " + nameof(HxInputBase<TValue>) + " has to be set.");
```

There is one more item than usually - _HxInputBase.Defaults?.ValidationMessageDisplayMode_ - bacause descandants have its own settings **and defaults** with nullable ValidationMessageDisplayMode and we need to apply the static one. But I cannot use the **virtual** GetDefaults() because it is overriden... To be discussed.

jirikanda:
I have implemented the "Blazor core", now it is your turn, the CSS squad (@crdo, @Harvey1214).

PR #165, opened 2022-05-20
UpdateContentRenderer performance optimization

hakenr:


hakenr:
merging but still good to review @jirikanda

Issue #166, opened 2022-05-20
[HxInputTags] Missing Label if LabelType="LabelType.Floating"


crdo:
[HxInputTags] Missing Label if LabelType="LabelType.Floating"

PR #179, opened 2022-05-25
[HxInputTags] Missing Label if LabelType="LabelType.Floating" #166

Harvey1214:


Havit.Blazor.Components.Web.Bootstrap/Tags/Internal/HxInputTagsInternal.razor
```diff
@@ -1,6 +1,6 @@
 @namespace Havit.Blazor.Components.Web.Bootstrap.Internal
 
-<span class="@CssClassHelper.Combine((HasInputGroupsEffective ? "input-group" : null), InputSizeEffective.AsInputGroupCssClass(), InputGroupCssClass)">
+<span class="@CssClassHelper.Combine((HasInputGroupsEffective ? "input-group" : null), InputSizeEffective.AsInputGroupCssClass(), InputGroupCssClass, "form-floating")">
```
hakenr, line 3: I doubt we want the `form-floating` class to be there all the time....?

Havit.Blazor.Components.Web.Bootstrap/Tags/Internal/HxInputTagsInternal.razor.cs
```diff
@@ -111,6 +111,11 @@ public partial class HxInputTagsInternal
 		/// </summary>
 		[Parameter] public RenderFragment InputGroupEndTemplate { get; set; }
 
+		/// <summary>
+		/// Label to render before input.
```
hakenr, line 115: The `Label` here is not rendered before the input, as far as I see. Should we say "Label (only rendered for LabelType.Floating. The wrapping HxInputTags renders the label on his own for other cases)"? Or is it used in other scenarios?

Havit.Blazor.Components.Web.Bootstrap/Tags/Internal/HxInputTagsInternal.razor.cs
```diff
@@ -122,6 +127,7 @@ public partial class HxInputTagsInternal
 		protected bool HasInputGroupsEffective => !String.IsNullOrWhiteSpace(InputGroupStartText) || !String.IsNullOrWhiteSpace(InputGroupEndText) || (InputGroupStartTemplate is not null) || (InputGroupEndTemplate is not null);
 
 		private string dropdownId = "hx" + Guid.NewGuid().ToString("N");
+		private string InputTagsControlId = "hx" + Guid.NewGuid().ToString("N");
```
hakenr, line 130: If it is private, it has to be lowerCase.

hakenr (changes requested):

Havit.Blazor.Components.Web.Bootstrap/Tags/Internal/HxInputTagsInternal.razor
Harvey1214: @hakenr I agree. This is still a work-in-progress - the label shows correctly but doesn't animate and fill the input when there is no value entered.

Harvey1214:
@crdo I've managed to get the floating label working when the HxInputTags component has some values, however, if the component is empty, the label stays in position and doesn't animate due to the current structure of the component.

I've thought of a solution but it's not ideal and the animation either wouldn't work or we would have to remake it.

I'd gladly appreciate any ideas for solutions to this issue, thanks :)

crdo:
@Harvey1214 I played around with it for a while and came to a conclusion that with the current design of floating labels in Bootstrap and HxInputTags in Havit.Blazor it would be very complicated to make this working.

PR #197, opened 2022-06-06
[HxSelect] Cannot be used for "add next" scenario #196

### LINKED ISSUE #196: [HxSelect] Cannot be used for "add next" scenario
* HxSelect should contain list of items to add (somewhere...)
* When the item is selected, it should be added (somewhere) and should be removed from the list.
* Handled by custom event handling (no databinding used)
* Value is always null and is not updated by custom event handler (because the selected item must be removed from the items and therefore cannot be used as a value)
* SelectValue should be changed to "null" value after user chooses item to add.

Right now, when the user picks the value it is removed from the list and the next one is used as selected items.
The reasons are
* using <option selected>
* not using SetUpdatesAttributeName

jirikanda:
* fixed #196

### DIFF
--- a/BlazorAppTest/Pages/HxSelectTest.razor
+++ b/BlazorAppTest/Pages/HxSelectTest.razor
@@ -2,18 +2,29 @@
 
 <h1>HxSelect</h1>
 
-<HxSelect TItem="decimal" TValue="decimal?" Data="@data" @bind-Value="@selectedValue" NullText="-all-" ValueSelector="(v => v)" />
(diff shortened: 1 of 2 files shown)


jirikanda:
* removed _builder.AddAttribute(..., "selected", ...)_
* added _builder.SetUpdatesAttributeName("value");_ (the only one component in the library uses this method)
* sequenceNumbers changes (therefore it looks are more changes)

hakenr (commented):
`Field 'HxSelectTest.selectedValue' is never assigned to, and will always have its default value`
...will fix on my own and publish the release.

Issue #218, opened 2022-06-24
[doc] Search - API section not updating on page change


Harvey1214:
Steps to reproduce

1. Go to any "type/*" page in the documentation.
2. Search for another type page.
3. Select the search result (click, enter, or arrows).
4. URL and title update, however, the API section stays unchanged.

An obvious solution would be to set ForceLoad=true when navigating to a page in the Search component, however, I find that undesirable since it would worsen the user experience.

cc @hakenr

PR #221, opened 2022-06-24
[doc] Search - API section not updating on page change #218

Harvey1214:


hakenr:
Already resolved within #214.

Issue #238, opened 2022-07-01
[HxAutosuggest] - HighlightFirstSuggestion - does not close dropdown when confirmed with Enter-key


hakenr:
1. doc - Search - type anything
2. hit Enter
3. the page changes, but the autosuggest dropdown remains open

...should be closed when the value is confirmed.

PR #241, opened 2022-07-07
[HxAutosuggest] - HighlightFirstSuggestion - does not close dropdown when confirmed with Enter-key #238

Harvey1214:


Havit.Blazor.Components.Web.Bootstrap/Forms/Autosuggests/Internal/HxAutosuggestInternal.razor.cs
```diff
@@ -225,6 +225,7 @@ private async Task HandleInputEnterKeyDown()
 			if (HighlightFirstSuggestionEffective)
 			{
 				await HandleItemClick(suggestions.FirstOrDefault());
+				await DestroyDropdownAsync();
```
hakenr, line 228: The event-callback can be long-running task, we should close the dropdown first and then call the callback?
@Harvey1214  Will fix on my own by exchanging those two lines (if you see any troubles with that, let me know).

hakenr (changes requested):

hakenr (approved):

Issue #224, opened 2022-06-27
[HxSidebar] Dropdown menu for section does not show up when sidebar is collapsed while section is expanded


alexax578:
When I expand a subsection in HxSidebar while sidebar is in expanded state:
![image](https://user-images.githubusercontent.com/37944469/175880555-b67e8329-aced-4747-8937-b18b883a56cf.png)

And then I collapse the sidebar while the section is expanded the dropdown menu that shows up for the section in collapsed state of the sidebar will never show up.
![image](https://user-images.githubusercontent.com/37944469/175881341-76e7c911-a5e3-4188-8827-1e5270cbdb43.png)

This behavior occurs after I repeat this process for the second time. First time when I do this the dropdown menu shows up.

Harvey1214:
@alexax578 I found more general repro-steps to this issue

1. Collapse the dropdown
2. Open a dropdown menu for a section
3. Close it
4. Expand, then collapse the sidebar
5. The section can't be opened

PR #245, opened 2022-07-08
#224 [HxSidebar] Fix subsections not opening in sidebar collapsed state

Harvey1214:
Fixed

1. Collapse the dropdown
2. Open a subsection (dropdown menu)
3. Close it
4. Expand, then collapse the sidebar
5. The section can't be opened

### DIFF
--- a/Havit.Blazor.Components.Web.Bootstrap/Forms/SearchBox/Internal/HxSearchBoxItem.razor
+++ b/Havit.Blazor.Components.Web.Bootstrap/Forms/SearchBox/Internal/HxSearchBoxItem.razor
@@ -14,7 +14,7 @@
 		}
 		else
 		{
-			<div class="hx-search-box-item-title">@Title</div>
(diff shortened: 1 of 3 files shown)


hakenr:
@crdo Please review, Is seems it does not target the original issue (reactivating dropdown menu when it reapears in DOM), but it resolves it by not removing the menu from the DOM at all.

hakenr (approved):
lgtm

Harvey1214:
@crdo @hakenr I understand this solution isn't ideal, therefore I have solved the issue using a different approach in [this commit](https://github.com/havit/Havit.Blazor/commit/9dbee084cc287aef97c348084ae3a82a8ece6a71).

It renders the whole HxDropdown only when the sidebar is collapsed. The item content (text, icon, ...) is saved in a field and used in both states (collapsed, expanded) since it doesn't change between them.

hakenr (approved):
lgtm

Issue #65, opened 2022-01-13
[HxTreeView] - Expand/collapse transition animation


hakenr:
Use `HxCollapse` for implementation?

Issue #182, opened 2022-05-27
[HxAccordionItems] are not collapsed correctly (not tracking DOM state?)


pepe-marba:
I found a problem when using the HxAccordion component, This Accordion displays a collection of items that need to be sorted as per user request.

These items have sub items that are displayed when the HxAccordionItem component is expanded.

The problem comes after sorting the items, some times there are more than one item expanded when it's supposed that only one should be displayed as expanded.

I have created a small project to test this, and the steps to reproduce the error are these:
1.- Expand any item, but the first one, in the Teams to Sort section.
2.- Click the Sort DESC button.
3.- Expand the first Item in the list
4.- Click the Sort ASC button.
5.- You should see two expanded items displayed.

These is another way to reproduce the issue:
1.- Expand the first item
2.- Click the Sort Desc button
3.- Expand the element in the first position after sorting the items.
4.- Two items should be displayed as expanded.

I found other different ways to reproduce the issue but I wasn't able to reproduce them in order to explain the steps to follow.

In this [link ](https://github.com/pepe-marba/havit-demo) you could find the small project I created for this.

This is how the 2 items look as expanded
![image](https://user-images.githubusercontent.com/8064119/170776031-e35be9d5-e175-41fa-ad78-184f8a700525.png)

Thanks

hakenr:
The current HxAccordion implementation does not track DOM state. When user expands/collapses any item, the Blazor-HxAccordionItem is not aware of it. When re-rendering, Blazor identifies and applies only some DOM differences and the reported behavior occurs.

The solution is to track DOM state in HxAccordion, or better to track DOM state in HxCollapse na and refactor HxAccordion to use HxCollapse internally (#79).

The only thing we are afraid od is the cost of DOM state tracking (listening to hidden/shown events). This might bring some performance costs. We will consider adding TrackState parameter to such components (HxDropdown is another example).

hakenr:
@pepe-marba We are still refactoring the HxAccordion and trying to make it better, but for now, you can resolve your issue by setting `@key="@team.Id"` for `HxAccordionItem`.

hakenr:
In the end, after refactoring HxAccordionItem to HxCollapse, we decided to go for `@key` requirement in dynamic scenarios. We might change this in future (won't be breaking change).

[...]

Issue #19, opened 2021-10-05
[HxDropdownToggle] Dynamic CssClass value leaves dropdown menu stuck in open state


tommi1hirvonen:
Having the `CssClass` property for `HxDropdownToggle` change dynamically based on user interaction, for example, leaves an open dropdown menu stuck in an open state. It looks like the class attribute `show` is overwritten for the dropdown toggle whenever the `CssClass` value changes.

Here is a simple working example. Toggling the checkbox inside the dropdown menu leaves the dropdown stuck in an open state. The class attribute `show` is removed from the dropdown toggle whenever the `IsBold` property in this example is changed.

```
<HxDropdown AutoClose="DropdownAutoClose.Outside">
    <HxDropdownToggle Color="ThemeColor.Primary" CssClass="@(IsBold ? "fw-bold" : null)">Show dropdown</HxDropdownToggle>
    <HxDropdownContent>
        <div class="row m-3" style="min-width: 100px;">
            <div class="col">
                <div>
                    <HxInputCheckbox @bind-Value="IsBold" Label="Toggle button text bold" />
                </div>
            </div>
        </div>
    </HxDropdownContent>
</HxDropdown>

@code {
    private bool IsBold { get; set; } = false;
}
```

It looks like the `CoreCssClass` property in `HxDropdownToggle.cs` doesn't take into account whether the dropdown is being shown or not. I guess this might the underlying reason.

```
protected override string CoreCssClass =>
	CssClassHelper.Combine(
		base.CoreCssClass,
		"dropdown-toggle",
		(DropdownContainer?.Split ?? false) ? "dropdown-toggle-split" : null,
		(NavContainer is not null) ? "nav-link" : null);
```

hakenr:
Hi @tommi1hirvonen,
I was able to reproduce your scenario in our dev environment and it seems it is related to the fact you pointed out - the disappearing `show` class on `HxDropdownToggle`.

The current implementation is kind of "basic" and does not cover such advanced scenario. The full implementation is currently only in `HxAccordion` (Blazor component listens to DOM events and tracks the shown/hidden state) and we will have to replicate such behavior in `HxDropdownToggle` and `HxCollapseToggleButton`.

(+ still considering renaming the `HxDropdownToggle` to `HxDropdownToggleButton` and freeing the `HxDropdownToggle` name for future non-button generic toggles.)

---

I will try to dig into this issue this week to be able to get the enhanced version into next release.

PR #246, opened 2022-07-08
[HxTreeView] - Expand/collapse transition animation #65

Harvey1214:


hakenr (changes requested):
Does not reflect item's `ItemInitialExpandedSelector`

Harvey1214:
@hakenr the item needs to be expanded initially if InitialExpandedSelector returns true on the Item. The question is, where should one expand the collapse? Please consider the following.

- Cannot call `Expand()` in `OnParametersSetAsync` because `collapseComponent` is null at that time (component not rendered yet - populated through `@ref`)
- Cannot call `Expand()` in `OnAfterRenderAsync(firstRender)` on `firstRender` because although `collapseComponent` is not `null`, it works not quite as one would expect - sometimes opens without issues, sometimes the `HxCollapse` doesn't do anything. It's quite random, sometimes it opens correctly 4 times in a row, the ratio is about `3/2` (`success / failure`).

The current solution is more of a work-around. Any ideas on a less "hacky" solution would be appreciated, thanks.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
```diff
@@ -42,19 +48,28 @@
...
-                        ContentTemplate="this.ContentTemplate"
-                        OnItemSelected="this.OnItemSelected" />
-    }
-}
\ No newline at end of file
+    <HxCollapse
+        @ref="collapseComponent"
+        CollapseDirection="CollapseDirection.Vertical"
+        OnShown="HandleAnimationCompleted"
+        OnHidden="HandleAnimationCompleted"
```
hakenr, line 57: I think we have to listen to these events and set `IsExpanded` accordingly. If we set `IsExpanded` only when on our expander-click, we can loose other signals where the collapse gets collapsed/expanded. There might not be such case now, but imagine future development where we might want to use accordion-mode of the `HxCollapse` (with `data-bs-parent;) where opening one node automatically closes another.

See `ShowAsync`, `HideAsync`, `HandleJsShown` and `HandleJsHidden` in `HxCollapse.razor.cs` for inspiration.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
```diff
@@ -2,13 +2,19 @@
...
+    IEnumerable<TItem> children = this.ChildrenSelector(this.Item);
+    bool hasChildren = children?.Any() ?? false;
+    bool isSelected = this.Item.Equals(this.TreeViewContainer.SelectedItem);
+    bool shouldBeInitiallyExpanded = this.InitialExpandedSelector?.Invoke(this.Item) ?? false;
+
+    if (!hasBeenInitiallyExpanded && shouldBeInitiallyExpanded && hasChildren)
+    {
+        hasBeenInitiallyExpanded = true;
+        IsExpanded = true;
+    }
```
hakenr, line 17: Not sure here.
The original implementation was too simple to follow.
The `InitialExpandedSelector` should be used just once in the lifecycle (`OnInitialized`?) and should not be called on every single re-render.

hakenr (changes requested):

Havit.Blazor.Components.Web.Bootstrap/Collapse/HxCollapse.razor
```diff
@@ -1,5 +1,5 @@
 ﻿@namespace Havit.Blazor.Components.Web.Bootstrap
 
-<div @ref="collapseHtmlElement" id="@this.Id" class="@GetCssClass()" data-bs-parent="@this.Parent" @attributes="this.AdditionalAttributes">
+<div @ref="collapseHtmlElement" aria-expanded="@InitiallyExpanded" id="@this.Id" class="@GetCssClass()" data-bs-parent="@this.Parent" @attributes="this.AdditionalAttributes">
```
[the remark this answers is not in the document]
Harvey1214: I left it this way because JS overrides the `aria-expanded` attribute anyways but I agree this doesn't look good and may cause issues, so I'll adjust it.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
Harvey1214: You're correct, however, this brings forth another issue -> the icon will rotate only once the collapse is expanded, which isn't pleasing to watch. Should I solve this issue now? It would probably require listening to the `show` and `hide` events on the collapse using JS in order to notify the `HxTreeViewItem` when the collapse starts animating.

Havit.Blazor.Components.Web.Bootstrap/Collapse/HxCollapse.razor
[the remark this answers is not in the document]
hakenr: I think JS changes the value when any ACTION in UI occurs (JS sets new attribute value), but if Blazor re-renders the component on its own, it won't be "fixed" by JS. See similar issues:
https://github.com/havit/Havit.Blazor/issues/19
https://github.com/havit/Havit.Blazor/issues/182

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
hakenr: Is there any chance to remove the `expanded` class on `hx-tree-view-item-expander` and to rely just on the `show` class being added by the `HxCollapse`?
It is added at the beginning of the animation. Accordion most likely uses this class to control the "arrow" direction (https://getbootstrap.com/docs/5.1/components/accordion/).

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
```diff
@@ -42,19 +41,28 @@
...
+                            Item="item"
+                            TitleSelector="this.TitleSelector"
+                            IconSelector="this.IconSelector"
+                            CssClass="@this.CssClass"
+                            CssClassSelector="this.CssClassSelector"
+                            ChildrenSelector="this.ChildrenSelector"
+                            Level="@(this.Level + 1)"
+                            ContentTemplate="this.ContentTemplate"
+                            OnItemSelected="this.OnItemSelected" />
+        }
```
hakenr, line 65: We should probably add `@key="item"` to the `HxTreeViewItemInternal` usages (incl. the top-level iteration) to guarantee the component won't be reused if the `item` changes (can bring new result of `InitiallyExpandedSelector` etc.)

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor.cs
```diff
@@ -18,16 +18,82 @@ public partial class HxTreeViewItemInternal<TItem> : ComponentBase
...
+			{
+				await Expand();
+			}
+		}
+
+		/// <summary>
+		/// Expand the item with an animation.
+		/// </summary>
+		/// <returns></returns>
+		public async Task Expand()
```
hakenr, line 59: The names should be `ExpandAsync` and `CollapseAsync`.
Do we want to expose these methods as `public`? Do we call them from outside?
(Let's keep the internal components as encapsulated as possible.)

hakenr (changes requested):

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
Harvey1214: At the beginning of the animation, the `collapsing` class is added, and the `show` class is added only once the animation is completed. During the animation, only the `collapsing` class is present, so we have no way to tell what direction the arrow should be facing.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor.cs
Harvey1214: I was thinking it may come in handy if the user wanted to for example create the functionality `InitiallyExpanded` provides but with the animation. I've made them `protected` for now, let me know if you find that ok.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
hakenr: How is this handled by Bootstrap accordion arrow? @crdo

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor.cs
hakenr: If we will want to use it, we will change it to anything more accessible later. Internal components are not meant to be used by any user in any scenario. They are only meant to be used in Havit.Blazor library itself. They are `public` only for the reason the `Component` has to be public to be able to use it in `.razor` markup.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
Harvey1214: @hakenr It looks for a `collapsed` class on the accordion button, not the collapse itself as shown below.
image

I'll solve it this way too by using `HxCollapseToggleButton` or `HxCollapseToggleElement`. A `collapsed` class is added or removed right at the start of the animation, exactly what we need.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor.cs
Harvey1214: Ok, will change it to `private`.

Havit.Blazor.Components.Web.Bootstrap/TreeViews/Internal/HxTreeViewItemInternal.razor
```diff
@@ -42,19 +39,30 @@
...
-                        OnItemSelected="this.OnItemSelected" />
-    }
-}
\ No newline at end of file
+    <HxCollapse
+        Id="@collapseId"
+        CollapseDirection="CollapseDirection.Vertical"
+        InitiallyExpanded="initiallyExpanded"
+        OnShown="HandleAnimationCompleted"
+        OnHidden="HandleAnimationCompleted">
```
Harvey1214, line 49: @hakenr `HxCollapse` closes or opens without animation when the `HxTreeViewItemInternal` isn't rerendered in between `HxCollapse` state changes. I haven't found the underlying issue yet.

Harvey1214: @hakenr could you please merge this branch, so that I can complete the other PR. I don't think this work-around causes any issues. Thanks

hakenr: Fixed the issue with StateHasChanged + merging.

hakenr (approved):
lgtm

PR #265, opened 2022-07-25
[HxGrid] OnDataItemClicked

Harvey1214:
Handling the OnClick event on grid items (rows) is crucial for features such as [this task](https://dev.azure.com/havit/DEV/_workitems/edit/63065/) on Goran - when we want to for example edit a grid item in an offcanvas or anywhere outside the grid, we need a way to find out if an item has been clicked.


hakenr:
Covered by `SelectedItem`

PR #268, opened 2022-07-27
[HxCollapse] listen to show and hide events

Harvey1214:
Necessary for BtSidebar where `OnShow`, all other items should be hidden.


hakenr:
`BtSidebar` resolved by introducing `MultipleItemsExpansion` parameter. Closing this PR as the proposed EventCallback do not align with the rest of the library (currently, we expose only on-transition-completed events, not the on-transition-start events). Can  be easily added any later if needed.

PR #286, opened 2022-08-09
[HxSidebar] Keep items expanded

Harvey1214:
Needed to solve [this bug](https://dev.azure.com/havit/DEV/_workitems/edit/63485/).


Harvey1214:
@hakenr can we move forward with this PR? I think the discussion in [AB#63485](https://dev.azure.com/havit/fd51f9bd-46eb-4a1b-98c8-b26cba5651c4/_workitems/edit/63485) is resolved and the agreed upon solution has been implemented.

hakenr:
(This PR = "má si pamatovat předchozí stav")

hakenr (changes requested):
@Harvey1214 Are you sure this PR is still relevant? The "Files changed" view seems chaotic to me:
* new `KeepItemsExpanded` parameter which is not being used et al,
* not sure regarding setting the `InitiallyExpanded` parameter of `HxCollapse` - is this the solution of "we will track/remember the state"?

Harvey1214:
@hakenr I forgot to remove the `KeepItemsExpanded` parameter, we decided that this functionality wouldn't be adjustable, that's why it's not used anywhere.

hakenr (approved):
lgtm

Issue #273, opened 2022-07-29
[HxSidebarItem] Add SidebarCollapsed to SidebarItemContentTemplateContext for ContentTemplate


hakenr:

Harvey1214:
@hakenr I don't see the purpose of this addition since `ContentTemplate` is not displayed when the sidebar is collapsed - the user would have to override our CSS to make this adjustment relevant, which doesn't seem like a very common scenario to me.

![image](https://user-images.githubusercontent.com/62853076/183939745-80580736-f432-43b4-9c3b-02dda14a7ae7.png)

![image](https://user-images.githubusercontent.com/62853076/183939902-a5686c4b-a717-42df-9509-36e520d65069.png)

I suggest creating a new context for the `ChildContent` of the `HxSidebarItem` and adding the `SidebarCollapsed` property there because the `ChildContent` is always rendered - when a sidebar is collapsed or expanded.

If you still think we should add this feature, please let me know, otherwise, I will implement the new context as described above.

Thanks and have a nice day 😀

hakenr:
@Harvey1214 Ahh, I see, the `ContentTemplate` is currently not visible if collapsed.
The proposed context for `ChildTemplate` is also weird, it duplicates the cascaded `ParentSidebar`.
Let's keep this unimplemented and we will see what demands will come from real projects.

PR #290, opened 2022-08-10
[HxSidebarItem] Add SidebarItemChildContentContext #273

Harvey1214:


hakenr:
@Harvey1214 Ahh, I see, the `ContentTemplate` is currently not visible if collapsed.
The proposed context for ChildTemplate is also weird, it duplicates the cascaded ParentSidebar.
Let's keep this unimplemented and we will see what demands will come from real projects.

Issue #260, opened 2022-07-16
[HxButton] Space between Icon and Text


crdo:
Currently, the gap between `HxButton` `Icon` and `Text` is made with `<text>&nbsp;</text>`.
That makes it impossible to configure and in case of `Color="ThemeColor.Link"`, the whitespace between `Icon` and `Text` is underlined.
I suggest to change the spacing to margin and make it configurable via CSS var.

PR #293, opened 2022-08-11
[HxButton] Space between Icon and Text #260

Harvey1214:
There are two classes (one for when the icon is before the text and for after the text), this was made to allow for easy customizations in case the user wanted differing margins for these two scenarios.


hakenr (approved):
lgtm, merging
@crdo, please take a look at it (adjust in master, if needed)

Issue #294, opened 2022-08-11
[HxInputDate] dropdown menu (calendar) can't be closed by clicking outside of it


Harvey1214:
1. click the input (gets focus, a dropdown menu is opened)
2. write an invalid date
3. press `enter` or click outside of the input - `OnChange` event is called
4. dropdown menu **cannot** be closed by clicking outside of it (only by selecting a date in the calendar, clicking the "Clear" button or the "OK" button)

@hakenr, @crdo this behavior seems like a bug, in case it is intentional, let me know and I'll revert the changes. Thanks and have a nice day.

hakenr:
@jirikanda Was it intentional? From my point of view, it makes sense to prevent closing the calendar when there is no valid date selected (or null). In such case we can reject the PR.

@Harvey1214 Anyway, if there is invalid input, the `Clear` button should remove the input (it does not) and the `OK` button should  probably keep the HxCalendar open. Currently, both will close the HxCalendar and leave the invalid input in the box.

jirikanda:
HxInputDate modified to use HxDropdown&co instead of "plain div with dropdown classes". The main reason to this changes is that HxDropdown tracks the state and renders "show" class to the dropdown trigger element (input here) whenewer the dropdown should be visible.

jirikanda:
I tried to modify the implementation of HxInputDateRange to work internally the same way as HxInputDateRange (#307). But I found the limitation mentioned in #308. To align the HxInputDate and HxInputDateRange implementations I need
* to resolve the #308 issue
* OR to design the bootstrap html structure another way (two HxDropdowns?)
* OR not to align the component implementations...

FYI, @crdo

Harvey1214:
@jirikanda I have used your new implementation of `HxInputDate` using Hx components and modified it so that it works in Bootstrap v5.2. The implementation of `HxInputDateRange` has been aligned with `HxInputDate`.

These changes are in the PR for the Bootstrap upgrade - https://github.com/havit/Havit.Blazor/pull/274. There is nothing left to do on this issue I think, it shall be marked complete once this branch is merged into master.

PR #295, opened 2022-08-11
[HxInputDate] dropdown menu (calendar) can't be closed by clicking outside of it #294

Harvey1214:


hakenr:
@jirikanda Please review, these were "your" components.
@Harvey1214 Should we add the same fix to `HxInputDateRangeInternal`?

Issue #288, opened 2022-08-09
OnToastHidden


pallu:
Would it be possible to call the OnToastHidden of HxToast from MessengerMessage or AddMessage?

hakenr:
@pallu Not sure what you are trying to achieve. Can you please give me more context?

pallu:
When a toast closes - either by clicking the x or on auto-hide, the OnToastHidden gets fired on HxToast. However, it's internal to the library. If you can expose it to the HxMessenger, then for an AddMessage, a callback can be passed to be fired when the toast is hidden.

Harvey1214:
Hello @pallu,

thank you for the context, now we understand what functionality you are requesting. We will consider implementing this feature.

Thank you for using our component library and have a nice day 😀

pallu:
Thank you!

pallu:
My specific application is that I keep a track of toast messages for history. However, if a toast is closed manually, I remove it from the list - I know that the callback really does not track if a toast is closed manually or auto-closes. Anyway, for now I use the MutationObserver in JS for added and removed toasts and I have been able to successfully use that for my application

hakenr:
@pallu I see. For such an application I would recommend to use `HxToast` instances directly (in `foreach` or `HxRepeater`) and bind them to a list of messages you will maintain in your state and manipulate as needed.

We were discussing the possibility to add `OnMessageClosed` to `HxMessenger`, but it opens much more questions than it resolves. For now, we decided not to add such feature to the library, unless there will be more scenarios/users demanding it.

If you get stuck in implementing your scenario with `HxToast`, place you question to the _Discussions_ here and we will try to help you.

Thanks for using our library and providing feedback...

PR #297, opened 2022-08-11
[HxMessenger] OnMessageClosed #288

Harvey1214:
I considered adding the `OnMessageClosed` event to the `IHxMessengerService`, however, I concluded that it would be quite chaotic to intermingle our internal logic for adding messages with events users may subscribe to and that's why I decided to simply add an `EventCallback` parameter to the `HxMessenger` component.

### DIFF
--- a/Havit.Blazor.Components.Web.Bootstrap/Toasts/HxMessenger.razor
+++ b/Havit.Blazor.Components.Web.Bootstrap/Toasts/HxMessenger.razor
@@ -5,13 +5,12 @@
         @foreach (MessengerMessage message in messages)
         {
             // @key is used to pair messages to HxToasts when some message disappers.
+            // We are not passing the Title and Text properties of the MessengerMessage to the HxToast because we are using them to build the ContentTemplate.
             <HxToast @key="@message.Key"
                         Color="@((message as BootstrapMessengerMessage)?.Color)"
                     CssClass="@message.CssClass"
                     HeaderIcon="@message.Icon"
-                    HeaderText="@message.Title"
                     HeaderTemplate="@message.HeaderTemplate"
-                    ContentText="@message.Text"
                     ContentTemplate="@message.ContentTemplate"
                     AutohideDelay="@message.AutohideDelay"
                     OnToastHidden="() => HandleToastHidden(message)" />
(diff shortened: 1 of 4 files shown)


Havit.Blazor.Components.Web.Bootstrap/Toasts/HxMessenger.razor
```diff
@@ -5,13 +5,12 @@
         @foreach (MessengerMessage message in messages)
         {
             // @key is used to pair messages to HxToasts when some message disappers.
+            // We are not passing the Title and Text properties of the MessengerMessage to the HxToast because we are using them to build the ContentTemplate.
```
hakenr, line 8: @Harvey1214 This is not true. It is just the extension-methods which use the `ContentTemplate` approach, but you can still pass the full `MessengerMessage` instance via `IHxMessengerService.AddMessage(MessengerMessage message)` and use the `Title` and `Text` properties of the `HxToast`.

hakenr (changes requested):
@Harvey1214 I'm still pretty unsure if this feature makes sense (adds value greater than the maintenance costs). Will discuss with the user first.

(We might add some meaningful demo than.)

Havit.Blazor.Components.Web.Bootstrap/Toasts/HxMessenger.razor.cs
```diff
@@ -11,6 +11,11 @@ public partial class HxMessenger : ComponentBase, IDisposable
 		/// </summary>
 		[Parameter] public ToastContainerPosition Position { get; set; } = ToastContainerPosition.None;
 
+		/// <summary>
+		/// Fires when a message (a toast) is hidden by close-button click or through auto-hide.
+		/// </summary>
+		[Parameter] public EventCallback<MessengerMessage> OnMessageClosed { get; set; }
```
hakenr, line 17: @Harvey1214 The `HxMessenger` is usualy sitting in `App.razor` or `MainLayout.razor`. I don't think you would want to listen to this callback there as the messages are added from the user-code via `IHxMessengerService` and that is the place where you "know" what you want to do when the message is "closed".

This just adds my doubts whether this whole feature is a good idea.

hakenr:
Rejecting the PR for the reasons described above.

Issue #191, opened 2022-06-01
[HxCheckbox] Replacement of [HxInputCheckbox] with `Text` instead of `Label`


hakenr:
Current `HxInputCheckbox` uses `Label` parameter for text of the checkbox and does not support `form-label`.
This turned to be counter-intuitive and thus we decided to consolidate the component and use `Text` parameter for the `form-check-label`-label and the `Label` parameter for the `form-label`.

BUT as this is major breaking change which changes meaning of the `Label` parameter, we decided to create a brand new component `HxCheckbox` with the new API and mark the original component as `[Obsolete]` + drop it in later major release.

cc @jirikanda; @crdo

Issue #312, opened 2022-08-24
[HxCheckbox] [HxSwitch] Inline mode does not work in EditForm


jirikanda:
The [demo](https://havit.blazor.eu/components/HxInputCheckbox#Inline) works well.
But when I place the inputs to the EditForm, the inline mode does not work (each checkbox is on a separate line).

@crdo, how should the html look to make it working?

jirikanda:
But we are going to replace the component, see #191.

jirikanda:
Also happed for the new HxCheckbox, we need deep-dive. @crdo could you help please.

On the first look it looks like there is difference in display (block vs. inline-block) and width attributes on the wrapping div.

crdo:
@jirikanda it has to be wrapped in extra div.

hakenr:
Discussed the details with @crdo, my rewording:

As our CSS defaults bring flex-column to `<form>` element (`EditForm` component), every single direct child component of `<form>` is stacked in vertical column. (We can discuss this separately, we decided to go for this setup for better developer experience and ease of building the forms.)

This setup implies, whenever you want to go for horizontal stacking, you have to wrap your items in some element (usually `<div>`), and this is exactly what happens with inline checkboxes and what is needed to arrange them horizontally.

Closing this issue as _as designed_ for now. (+ Will add some explanation and/or guidance to documentation.)

hakenr:
> Is there a demo to show how to use the HxCheckbox inside and outside of the EditForm?

https://havit.blazor.eu/components/Inputs#form-layout - demo
https://havit.blazor.eu/components/HxCheckbox#Inline - alert

PR #274, opened 2022-07-29
Bootstrap 5.2

hakenr:

PR #314, opened 2022-08-24
[HxCheckbox] Replacement of [HxInputCheckbox] with Text instead of La…

jirikanda:
Issue #191.


hakenr:
`Inline`
* use the `form-check-inline` to parent (wrapping) `div`
* doc: not usable with `Label` + validation messages

jirikanda:
Inline mode implemented, but there is the issue #312.

hakenr:
@jirikanda 
Trying to finish the PR and found we forgot to replace `HxInputSwitch` with `HxSwitch` (some issue `Label` vs `Text`).
Will merge this PR (needed for #274) and elaborate on `HxSwitch` separately.

Issue #289, opened 2022-08-10
[HxDropdown] JavaScript interop calls cannot be issued at this time. This is because the circuit has disconnected and is being disposed


rqx110:
Step reproduce:
1. add a demo DropDown sample https://havit.blazor.eu/components/HxDropdown#basic-usage to a page
2. run blazor project (I am using balzor server type project)
3. navigate to the page (http://localhost:5298/)
4. refresh this page again, you will see below error message:
![image](https://user-images.githubusercontent.com/1829209/183846419-16f18d81-fb46-4ac1-b2c1-c66405fc906e.png)

Harvey1214:
Hello @rqx110,

this happens when the project hasn't been properly set up. For the dropdown to work, you have to set up the project according to our [Getting started](https://havit.blazor.eu/) page - include Bootstrap CSS, JS, register our services, etc...

If you've done this and the dropdown still doesn't work, then let us know and we'll look into it further.

Have a great day 😀

rqx110:
see, i have setup it:
![image](https://user-images.githubusercontent.com/1829209/184045459-6c989c9c-879e-4171-8f07-2d03fb38724d.png)
![image](https://user-images.githubusercontent.com/1829209/184045490-fa7514eb-d894-4329-8cd7-8e2d1a1dd99f.png)

rqx110:
**It only happend on .net6 target**

hakenr:
It seems (https://github.com/dotnet/aspnetcore/issues/33336) we have to guard all JS-calls in disposal logic with `try-catch(JSDisconectedException)`.

rqx110:
@hakenr 
`await jsModule.DisposeAsync();` may also cause `JSDisconectedException` while refreshing the page
I have tested below component will cause exception:
- `HxCarousel`
- `HxCollapse`
- `HxDropdownToggleButton`
- `HxDropdownToggleElement`
- `HxInputDateInternal`
- `HxInputDateRangeInternal`
- `HxModal`
- `HxScrollspy`
- `HxOffcanvas`
- `HxToast`
- `HxTooltipInternalBase`

hakenr:
@rqx110 Thanx for the contribution!

PR #319, opened 2022-08-25
fix JSDisconectedException on .NET6

rqx110:
relate https://github.com/havit/Havit.Blazor/issues/289#issuecomment-1217369424


hakenr (approved):
lgtm

Issue #303, opened 2022-08-16
[HxGrid] InfiniteScroll + synchronous DataProvider returning empty data => DOMException: Failed to execute 'setStartAfter' on 'Range': the given Node has no parent.


hakenr:
If there is HxGrid with ContentNavigationMode="GridContentNavigationMode.InfiniteScroll" (Virtualize) and the DataProvider delagate returns empty data synchronously, the underlying Virtualize emits JavaScript error to the console:
![image](https://user-images.githubusercontent.com/12828888/184900047-668bbf3e-f6f0-4c8d-9732-ead26ea4d435.png)
```
blazor.webassembly.j…1pV1-b5sQ3Rj11QGI:1 
 Uncaught DOMException: Failed to execute 'setStartAfter' on 'Range': the given Node has no parent.
    at https://localhost:44376/_framework/blazor.webassembly.js?v=wC-zAyYHVTNzevCw3SFvHI4jG51pV1-b5sQ3Rj11QGI:1:28726
    at Array.forEach (<anonymous>)
    at IntersectionObserver.<anonymous> (https://localhost:44376/_framework/blazor.webassembly.js?v=wC-zAyYHVTNzevCw3SFvHI4jG51pV1-b5sQ3Rj11QGI:1:28677)
(anonymous)	@	blazor.webassembly.j…1pV1-b5sQ3Rj11QGI:1
(anonymous)	@	blazor.webassembly.j…1pV1-b5sQ3Rj11QGI:1
```

In InfiniteScroll mode, we set `shouldRenderData` to true when `totalCount == null` to render the `Virtualize` which initiates the `DataProvider` request. As the DataProvider returns empty data, we re-render with `shouldRenderData = false` and remove the `Virtualize` from DOM. It seems it disappears from the DOM too quickly and the supporting JavaScript (MSFT) logs the error to console (not causing the Blazor app to fail!).

Related [AB#63587](https://dev.azure.com/havit/fd51f9bd-46eb-4a1b-98c8-b26cba5651c4/_workitems/edit/63587).

Workaround (for the scenarios where the Data is available in advance):
```razor
<HxGrid ContentNavigationMode="Data.Any() ? GridContentNavigationMode.InfiniteScroll : GridContentNavigationMode.Pagination" .../>
```

cc @alexax578 @jirikanda

jirikanda:
@hakenr Don't ask me why it helps ;-).

jirikanda:
[...]

PR #321, opened 2022-08-26
[HxGrid] InfiniteScroll + synchronous DataProvider returning empty da…

jirikanda:
…ta => DOMException: Failed to execute 'setStartAfter' on 'Range': the given Node has no parent. #303 (alternative approach)


hakenr (approved):
lgtm

Issue #322, opened 2022-08-26
[HxGrid] Columns should not allow sorting when there are no data


jirikanda:
Sorting forces data provider to read data from data source.
It is useless when we do not expect new data to arise.

Not sure if you agree, feel free to close the issue without PR merge.

PR #323, opened 2022-08-26
[HxGrid] Columns should not allow sorting when there are no data #322

jirikanda:
Issue #322.


hakenr (approved):
didn't like the `sortingEnabled` name
was not able to find better one
thus inlined the expression to prevent confusion
:-)

PR #328, opened 2022-08-30
doc - HxInputTextArea - don't display link to documentation page on the documentation page

Harvey1214:
This link is intended for IntelliSense.


hakenr (approved):
It is far from being ideal to rely on a bug when we "hide" the last line of documentation when it ends with link, but for now - why not.

Issue #335, opened 2022-08-31
[docs] Opposite logic for StaticBackdrop in Modal/Offcanvas demos


crdo:
We have opposite examples of `StaticBackdrop` for `Modal/Offcanvas` demos. Please use set or not set for both demos.
![image](https://user-images.githubusercontent.com/1410922/187759614-5aa3fba8-5a1f-4fec-afe1-f10dfb783838.png)
![image](https://user-images.githubusercontent.com/1410922/187759743-e167a5f4-10da-40da-ba12-00452d666b9c.png)

hakenr:
Wow. It is not the demo, it is the implementation in `HxModal`, which is wrong (in the light of today's library concept).

`HxModal` default for backdrop is `static` whereas Bootstrap has `true` as default.
We allow the `true`-mode by setting `UseStaticBackdrop="false"` parameter.
Bootstrap modal has third state `false` which is currently not supported by our `HxModal` (something we have to fix).

- [ ] We have to replace the `UseStaticBackdrop` parameter with full-featured enum-base `Backdrop` parameter with the same options as `OffcanvasBackdrop` (`True`, `False` and `Static`).
- [ ] The current default of `HxModal` is `static`. We have to decide whether we want to override the Bootstrap's `true` default with our `ModalSettings.Backdrop = ModalBackdrop.Static` or whether we stick with `ModalBackdrop.True` and let the users override the default if desired. The `true` option aligns with our strategy to "keep as much Bootstrap terminology and defaults as possible" the `static` option aligns with typical usage of modal behavior - a dialog which has to be resolved explicitly.
- [ ] We should decide whether we want to align default setting of `HxModal.Backdrop` and `HxOffcanvas.Backdrop`. Bootstrap aligns them to `true`. If we go for `HxModal.Backdrop = ModalBackdrop.Static` we might also want to se `HxOffcanvas.Backdrop = OffcanvasBackdrop.Static`.
- [ ] OPTIONAL: Alternative approach to defaults might be to add some setup-method (for `Startup.cs`) which would allow optional setup of "our applications" defaults (on top of those Bootstrap default which we would keep as library-defaults). Something like:
```csharp
services.AddHxServices().WithHavitApplicationDefaults();
```
 @crdo? @jirikanda?

PR #338, opened 2022-09-01
[HxModal] Fix Backdrop implementation, missing support for "false", not aligned with HxOffcanvas, changes Bootstrap default #335

Harvey1214:


hakenr (approved):
lgtm

Issue #341, opened 2022-09-01
[HxSidebar] Toggle button is rendered even in the navbar (mobile) version


Harvey1214:
1. Resize the browser window - decrease the width.
2. Click the hamburger icon to expand the sidebar.
3. Sidebar collapse toggle button is rendered on hover.
4. Click the button.
5. See the image below for result, which shouldn't be possible to achieve - the sidebar collapse toggle button shouldn't be visible in mobile view.

![image](https://user-images.githubusercontent.com/62853076/187956407-078036f6-9acc-4021-8ee9-c4e586a3a76c.png)

Happens also in our documentation's sidebar.

![image](https://user-images.githubusercontent.com/62853076/187958915-9cbd125b-120a-49cb-bf2b-a730b35212b9.png)

cc @crdo @hakenr

PR #342, opened 2022-09-01
[HxSidebar] sidebar toggle button is rendered even in the navbar (mobile) version #341

Harvey1214:


Harvey1214:
> When a user resizes the browser window (decreasing its width below the breakpoint) while the sidebar is collapsed, they won't be able to expand it horizontally because the toggle button has disappeared.

@crdo at the moment I cannot think of an elegant way to do this, therefore I suggest watching the toggle button with JS, and when it gets `display='none'`, we call a method from JS on the `HxSidebar` that will horizontally expand the sidebar.

Please let me know what you think, any ideas and/or feedback on this approach would be appreciated, thanks.

crdo (approved):

Harvey1214:
@hakenr, @crdo and I have concluded that the issue described above isn't a common scenario and not worth fixing, therefore, the current solution will be sufficient.

Issue #355, opened 2022-09-05
[HxInputDate] [HxInputDateRange] Move predefined dates to dropdown-menu


crdo:
I suggest to move the predefined dates UI to the HxInputDate[Range] dropdown-menu to make to UI more subtle and allow use of InputGroups.

PR coming.

PR #356, opened 2022-09-05
#355 [HxInputDate[Range]] Move predefined dates to dropdown-menu

crdo:
image

### DIFF
--- a/Havit.Blazor.Components.Web.Bootstrap.Documentation/Pages/Components/HxInputDateDoc/HxInputDate_Demo_CalendarButtons.razor
+++ b/Havit.Blazor.Components.Web.Bootstrap.Documentation/Pages/Components/HxInputDateDoc/HxInputDate_Demo_CalendarButtons.razor
@@ -1,5 +1,0 @@
-﻿<HxInputDate Label="Date" @bind-Value="@value" ShowCalendarButtons="false" />
-
-@code {
-    private DateTime value { get; set; }
-}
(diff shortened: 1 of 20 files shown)


hakenr:
## TODOs
- [ ] Do not preserve space for three-dots input-group when positioning the Icon (see below) @crdo 
- [ ] Consider `justify-content-between` for predefined dates/ranges (see below) @crdo
- [ ] Create new `ShowClearButton` parameter + remap the original `ShowCalendarButtons` to it and mark as `[Obsolete]` @hakenr 
- [ ] Update _Buttons in calendar_ demo @hakenr 

## Do not preserve space for three-dots input-group when positioning the Icon
image

## Consider `justify-content-between` for predefined dates/ranges
image

crdo:
@hakenr Changed the layout of `hx-input-date-dropdown-buttons` so when there is only one button, it is aligned to the left.
When there are two buttons there is `space-between` and when there are more than two buttons they are aligned to the left.

I also played around with different color and position of Clear button, but I don't like it...

image
image
image

crdo:
... the only way I like is when the Clear and Today would be on one line and the pre-defined days would be on a separate line. But since the Today is one of the pre-defined dates, it does not make sense...

hakenr:
> ... the only way I like is when the Clear and Today would be on one line and the pre-defined days would be on a separate line. But since the Today is one of the pre-defined dates, it does not make sense...

One of my thoughts was to align the predefined options to the right and wrap them "one line per item". Did you try something like that?

hakenr (approved):
lgtm, will merge to master

crdo:
@hakenr what about moving the clear button to the input itself same as we have it in HxAutosuggest so we can leave out the bottom of the dropdown for the predefined dates/ranges only?

crdo:
... the logic would be the same - if there is a value selected we show the clear button if the input is empty we show the calendar icon.

hakenr:
I think it will collide with the `CalendarIcon` or do you intend to "replace" it when focused?

crdo:
Yes, the same way we do it with HxAutosuggest. It makes sense to me.

PR #368, opened 2022-09-09
[HxAutosuggest][HxInputTags]: fix null reference exception

rqx110:
problem reproduce:
1. goto https://havit.blazor.eu/components/HxInputTags
2. in `Basic usage` demo, do not enter anything, press down key, error will occur

same step with `HxAutosuggest`.

this PR fix above issues


hakenr (approved):
lgtm, thanks for your contribution

Issue #365, opened 2022-09-08
[HxSearchBox] Clear text in input when abandoning the input and AllowTextQuery="false"


hakenr:
`AllowTextQuery="false`" - suggestions only (~ we should nearly align with `HxAutosuggest`)
1. Type anything in the box
2. Click anywhere outside the dropdown and box to close the dropdown.
...the text typed in the box should be cleared, but it stays there (should stay there only for `AllowTextQuery="false"`

(If `AllowTextQuery="true"` the text remains in the input, until some action is triggered. I.e. cleared at `OnItemSelected` or `OnTextQueryTriggered`)

cc @Harvey1214

PR #376, opened 2022-09-14
[HxSearchBox] Clear text in input when abandoning the input, AllowTextQuery="false" and suggestions dropdown menu doesn't have focus #365

Harvey1214:


hakenr (commented):
The event-handling method names should follow project conventions (`HandleXy`).
`OnXy` is not a suitable method name.
Method names should reflect the originating event (input vs. form etc.)
...will fix on my own and merge.

PR #379, opened 2022-09-14
[HxSearchBox] TextQuery item can be selected even when AllowTextQuery=false

Harvey1214:


Harvey1214:
@crdo @hakenr another question is whether we want to disallow selecting the input with the arrow keys when `AllowTextQuery=false` as it doesn't carry any purpose I believe (same thing for `HxAutosuggest`). Please let me know what you think, thanks.

hakenr (changes requested):
@Harvey1214 Same issue applies in the opposite direction - the start (for those scenarios, where we apply the "hightlight first suggestion" approach). We should keep the first item selected (should not "select the input"), when the `AllowTextQuery` is `false`.

hakenr (changes requested):
Both `HxAutosuggest` and `HxSearchBox` (SuggestionsOnly-mode) allow me to navigate "up" from the first item and get to the "nothing selected" state.
...which si something we are trying to solve, isn't it?

hakenr (approved):
lgtm

PR #380, opened 2022-09-14
[docs] On this page navigation

Harvey1214:


hakenr:
I will consolidate the content (and maybe add support for content levels, such as h2-h3-h4...).

@crdo can you please check the visual? I would prefer to add some responsiveness - make the basic content container narrower only if the navigation does not fit to the screen, but for larger screens (incl. FullHD), keep the original width of the container and stick the navigation to the right.
BTW: Is there any chance to make the navigation sticky (vertically)?

Harvey1214:
@hakenr I think I implemented both of the features you mentioned - multiple content levels and responsive design - on mobile, it is hidden in a collapse, on larger screens, it is in page, and on large screens, it is in a container on the right side of the screen similarly to what Bootstrap has on their documentation.

I'll look into it further tomorrow and check if anything doesn't work.

Harvey1214:
@hakenr currently, there are 3 different layout states which the `OnThisPageNavigation` component can be in. These are

1. sticky navigation on the right
image
2. navigation in page at the start of the page
image
3. navigation hidden in a collapse
image

Regarding the content levels, I think it is apparent from the screenshot that these are implemented (unlimited levels - these are set on the `SectionTitle` component).

I think all the points brought forward in your comment have been addressed, if you think of anything else, then let me know.

hakenr:
@crdo Can you please fine-tune the visual/UX?

The current responsivity behavior seems to be taken from https://getbootstrap.com/docs/5.2/getting-started/introduction/
Alternative: https://learn.microsoft.com/en-us/dotnet/api/system.diagnostics.stopwatch?view=net-7.0
Both work for me.

@Harvey1214 I would like to see "API" section of the page in the navigation. Only those headings applicable, eg.
```
Basic usage
Reverse
API
  Parameters
  Methods
  CSS Variables
```

Harvey1214:
Commit https://github.com/havit/Havit.Blazor/pull/380/commits/0b661d63541216c561547fbf6c16bdba5ad5c882 enables scenarios such as this one where we need to go down multiple levels at one time - go from level `3` to level `0`.

image

hakenr:
regarding multiple components

my opinion (@crdo can come up with some other visual idea)

let's consider `h1` as top-level (0?)
if there is only one top-level title on the page, don't render top-level to the navigation and shift all the levels one level up

eg.
```
HxButton
  A
  B
  C
  API
    Parameters
    Static properties
HxSubmit
  A
  B
  API
    Parameters
    Static properties
```
in comparison to single-component (single top-level) page:
```
A
B
API
  Parameters
  Static properties
```

hakenr (approved):
lgtm

PR #392, opened 2022-09-23
HxRadioButtonList - form-check as label

hakenr:
See #AB64636 work-item.
@crdo, prototyped an idea with @alexax578.

Is this valid (input in label; label in label; ...)?
Do we want to go this direction? Another solution for the original request?


alexax578:
@hakenr I believe that the issue we were trying to solve -> clicking on the form-check div triggers checkbox/radio button, has a documented solution in Bootstrap.
image
https://getbootstrap.com/docs/5.2/components/list-group/#checkboxes-and-radios

Making this draft obsolete.

hakenr:
Closing as obsolete.

PR #395, opened 2022-09-29
[HxMultiSelect] Text customizability

Harvey1214:


hakenr (changes requested):
Not sure about the `Text` name, but still, we should provide something like `Func<IEnumerable<TItem>, string> TextSelector` to be able to let the user assemble his own text.

Harvey1214:
I think the user will usually bind the value of the `HxMultiSelect` to a field they will have access to, therefore I don't find adding a `TextSelector` beneficial since they can assemble the `Text` using the values from the field that is bound to the component's value.

If you find such a selector helpful, then let me know.

hakenr (approved):
lgtm

Issue #466, opened 2023-02-09
[HxButton] To be able to set different icon color than Button text


crdo:
There is no easy way to make an icon different color than button text.

We could
- Add IconCssClass
- Add CSS var that would apply color on HxIcon (default unset)

PR #472, opened 2023-02-17
[HxButton] To be able to set different icon color than Button text #466

Harvey1214:


rqx110:
In this way, how we can set different icon color for every button?

rqx110:
hi @Harvey1214 , it is better provide a sample in doc to show how to change the icon color.

And, it looks like you have forgot to modify when icon in end placement:
https://github.com/havit/Havit.Blazor/blob/e33210a5cc0a6906c6b2e6fefeb7d68d62c6882f/Havit.Blazor.Components.Web.Bootstrap/Buttons/HxButton.razor#L51

Harvey1214:
@rqx110 I agree, the user will usually want to set the color to one of the theme colors, therefore the approach with the `IconCssClass` parameter is more suitable. Thank you for the recommendation.

Havit.Blazor.Components.Web.Bootstrap/wwwroot/defaults.css
```diff
@@ -8,6 +8,7 @@
 	--hx-autosuggest-dropdown-menu-height:				300px;	
 	/* HxButton */
 	--hx-button-space-between-text-icon:				0.25rem;
+	--hx-button-icon-color:								unset;
```
rqx110, line 11: not used any more. remove this in doc also

hakenr (approved):
lgtm

PR #473, opened 2023-02-20
[docs] HxGrid - individual column searching

Harvey1214:


hakenr:
TODOs:
* [ ] positioning of the search icon (cc  @crdo)
image

hakenr (changes requested):
see TODOs

Harvey1214:
@hakenr @crdo, I propose we allow the user to place the sort icon anywhere in the `HeaderTemplate`. I believe this can be achieved by introducing a `GridHeaderTemplateContext` with a `RenderFragment SortIcon` property. That way, the user can render the icon in the desired place in the `HeaderTemplate`.

PR #478, opened 2023-02-27
[HxAutosuggest] Remove need of HasAnyInputGroupEnd

crdo:
HasAnyInputGroupEnd and its logic adding "rounded-end" utility class inside HxAutosuggest structure caused issues with overriding border radiuses on inputs. I tried to get rid of it in this PR.
Tested validation, InputGroupAddons Start/End, all working.


hakenr (approved):
lgtm

PR #493, opened 2023-03-06
Fixes #492 [HxInputBase] form-label class should not be applied if LabelType.Floating

### LINKED ISSUE #492: [HxInputBase] form-label class should not be applied if LabelType.Floating
Currently, the CoreLabelCssClass is always "form-label" but for  LabelType.Floating [there should be no css class on the label](https://getbootstrap.com/docs/5.2/forms/floating-labels/#example).

crdo:
Fixes #492


hakenr (approved):
lgtm

PR #513, opened 2023-03-23
[HxGrid] Load more pagination loads increasingly larger page instead of additional pages

hakenr:
@jirikanda Please review.


hakenr:
reviewed with @jirikanda

PR #533, opened 2023-05-05
[HxModal] New event "OnHiding" with ability to cancel the hiding.

CryptKat:
New event "OnHiding" with ability to cancel the hiding.

Usage:
```razor
<HxModal OnHiding="OnModalHidingAsync">
    ...
    <HxButton OnClick="SubmitAsync"/>
</HxModal>

@code {
  private async Task OnModalHidingAsync(ModalHidingEventArgs args)
  {
    if (model.IsChanged && !(await MessageBox.ConfirmAsync(...)))
      args.Cancel = true;
  }
  
  private async Task SubmitAsync()
  {
    MyModelService.Save(model);
    model.IsChanged = false;
    await HideAsync(...);
  }
}
```

### DIFF
--- a/Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
+++ b/Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
@@ -1,10 +1,11 @@
-﻿export function show(element, hxModalDotnetObjectReference, closeOnEscape) {
+﻿export function show(element, hxModalDotnetObjectReference, closeOnEscape, subscribeToHideEvent) {
 	if (!element) {
 		return;
 	}
 
 	element.hxModalDotnetObjectReference = hxModalDotnetObjectReference;
-	element.addEventListener('hide.bs.modal', handleModalHide);
+	if (subscribeToHideEvent)
+		element.addEventListener('hide.bs.modal', handleModalHide);
 	element.addEventListener('hidden.bs.modal', handleModalHidden);
 	element.addEventListener('shown.bs.modal', handleModalShown);
 
--- a/Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
+++ b/Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
@@ -4,6 +4,7 @@
 	}
 
 	element.hxModalDotnetObjectReference = hxModalDotnetObjectReference;
+	element.addEventListener('hide.bs.modal', handleModalHide);
 	element.addEventListener('hidden.bs.modal', handleModalHidden);
 	element.addEventListener('shown.bs.modal', handleModalShown);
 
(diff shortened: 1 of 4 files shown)


hakenr (commented):
@CryptKat Thanks for your contribution and sorry for not getting to the review earlier.

It all makes sense, I'm just wondering if there could be some performance optimization for dominant use (without `OnHiding` callback), where no calls from JavaScript to Blazor and back would be made every time `handleModalHide` is raised. It might be sufficient to condition the entire handler on the presence of the callback (`OnHiding.HasDelegate`), but this would lose the ability to intercept the event in `override InvokeOnHidingAsync()`.

Any ideas how to make this conditional?

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
```diff
@@ -26,6 +27,21 @@ function handleModalShown(event) {
...
+    let modalInstance = bootstrap.Modal.getInstance(event.target);
+
+    if (modalInstance.dontPrevent)
+        return;
+
+    event.preventDefault();
+
+    let cancel = await event.target.hxModalDotnetObjectReference.invokeMethodAsync('HxModal_HandleModalHide');
+    if (!cancel) {
+        modalInstance.dontPrevent = true;
```
hakenr, line 40: 1. Is there any chance the `modelInstance` will be reused and the `dontPrevent` remains `true`? Should we set the attribute to `false` somewhere (for sure)?

2. I would go for some name which will contain the `hide` event name, such as `hidePreventionDisabled`.

hakenr (changes requested):

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
CryptKat: 1. As long as `hidden.bs.modal` event is triggered, modal instance is being disposed (inside `handleModalHidden` event handler). I tried, but couldn't get into a situation where modal instance is reused. 
2. Changed.
3. Accidentally I've seen an incorrect event listener removal. Fixed in 0527282.

hakenr: To accept such feature to the library I think we need the fast-path to be available for the major use case where the `hide` event is not intercepted. (As such intercept not only sends the signal to Blazor but also relatively aggressively influences the Bootstrap hiding workflow  - cancels it and "restores" it only if confirmed by Blazor-code).

The only way to implement it without incorporating any additional complex logic or new parameters is to rely on `OnHiding.HasDelegate` and sacrifice the `virtual InvokeOnHidingAsync()` seam allowing further logic in derived components.

Would you please add such logic to your PR? The main point is "if there is no `OnHiding` callback subscribed, do not subscribe to the `hide` event of the underlying Bootstrap component et al". (Will need some related logic to cover the conditional un-subscription in JS. You don't need to support the scenario where the `OnHiding` callback is set after the component is rendered for the first time.)

hakenr (changes requested):
see comment added above

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxModal.js
CryptKat: Conditional subscription to the "hide" event: ed3aee3

I didn't add conditional un-subscription, because of
> Calling removeEventListener() with arguments that do not identify any currently registered event listener on the EventTarget has no effect.

from https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

hakenr: We still need to remove the `protected virtual Task InvokeOnHidingAsync()` as we do not want to allow confusion for the authors of derived components to spend their time investigating why their `override InvokeOnHidingAsync()` is not being called (when the `OnHiding` parameter handler is not set). Just call directly `OnHiding.InvokeAsync()` from `HandleModalHide()`.

CryptKat: Done. 8b6bb17

hakenr (approved):
lgtm

hakenr:
@CryptKat Many thanx for your contribution!

PR #540, opened 2023-05-15
 [HxInputFileDropZone] Implement cascading enabled pattern

pollardpj:
In some scenarios it may be useful to disable HxInputFileDropZone inputs whilst doing background checks/custom posts/general state management of a form, where allowing a user to re-select files would not be appropriate.

This PR implements the common cascading approach (including respecting HxFormState).


hakenr (approved):
lgtm
Many thanks for your contribution @pollardpj!

PR #547, opened 2023-05-29
To add Animate parameter to HxModal to be able to remove fade animation.

crdo:
Currently we do not have a possibility to remove animation of HxModal.

https://getbootstrap.com/docs/5.3/components/modal/#remove-animation

This PR adds Animate bool to be able to remove the animation. Default true, no need for defaults/settings IMO.


hakenr (approved):
renamed to `Animated` (also considered `Fade`, but `Animated` aligns with other components)
added Settings/Defaults support
...lgtm

PR #557, opened 2023-06-13
[HxSearchBox] InputCssClass is not applied to the input when set

### LINKED ISSUE #556: [HxSearchBox] InputCssClass is not applied to the input when set
When I set some value to the Parameter InputCssClass on HxSearchBox, it is not applied to the HTML input like I would expect.

alexax578:
Resolves #556


hakenr (approved):
nice catch :-D

PR #566, opened 2023-07-17
To add selectors to HxAutosuggest icons

### LINKED ISSUE #565: [HxAutosuggest] differentiate between search icon and clear icon using CSS classes
I would like to be able to write CSS selectors that change colors of the icons based on whether the search icon is shown or the clear icon is shown. I don't want to target CSS classes of the icons themselves because they can be changed using the SearchIcon and ClearIcon parameters.

crdo:
Resolves #565 [HxAutosuggest] differentiate between search icon and clear icon using CSS classes


hakenr (approved):
lgtm

PR #573, opened 2023-08-03
HxAutossugest auto-scroll with arrow key navigation

hakenr:


hakenr (commented):
lgtm;
@crdo please review

PR #598, opened 2023-09-05
fixes #597 [HxCalendar] Add today parameter

### LINKED ISSUE #597: [HxCalendar] Add TimeZone support
This issue only affects Blazor Server.  But when the server is in a different timezone to the client then the Today highlight is on the wrong day.

I propose that any references to DateTime.Today/DateTime.Now either be converted using a supplied timezone or callbacks provided to get the today values.

Happy to provide a PR to solve this :)

blanchardglen:
This allows today to be set to client today in blazor server environments


blanchardglen:
Changing to use TimeProvider as per issue comments

Issue #607, opened 2023-09-19
[HxSearchBox] Scrolling when navigating with arrows


alexax578:
When I navigate using arrows in the dropdown menu I want the dropdown to automatically scroll to the currently focused item.

PR #608, opened 2023-09-19
Implement scrolling to focused element when using keyboard navigation on HxSearchBox

alexax578:
Resolves issue #607


hakenr (approved):
lgtm

Issue #616, opened 2023-10-05
[HxMultiSelect] - Filtering and/or searching


bholbrook:
The current HxMultiSelect is great but when the list contains a large number of items it can be difficult or slow to scroll and find the specific item you're looking to select.
If the component could be modified to allow typing to filter the displayed items, I think it would greatly improve the flexibility and usefulness of the component.

Perhaps something like how the Bootstrap 5 DataList functions, but with a multi-select: https://getbootstrap.com/docs/5.0/forms/form-control/#datalists

hakenr:
(This is probably for `HxAutosuggest` to support multiple items selection.)

bholbrook:
Would it be worth modifying the existing HxAutosuggest noting that we'd likely need a few different new multi-select properties just for this new functionality, or would it be better to create a new HxMultiSelectAutoSuggest?

hakenr:
Probably a new component derived from HxAutossuggest would be needed.
Anyways thanks for the feature request. We already heard that request from other users.
Currently, it is not on our list of priorities, but will keep it in the backlog.

bholbrook:
I've been attempting to build a new HxMultiSelectAutosuggest component utilising the core logic from the HxAutosuggest and have been making some progress but for at least the way I'm envisioning this component working I think it does make more sense to be something like a HxFilterMultiSelect.

My requirements and what I'm envisioning is a HxMultiSelect where the list is pre-populated with a pre-defined list of every available option, but then when you click on the dropdown and start typing it will start filtering down the available options in the dropdown list.
You'd be able to click or keyboard move through the items to select, and the filter would be removable through a clear X cross on the right-hand side of the input.
Once you clear the input the selected item text values will appear as a comma-delimited string, but regardless of the clear they'd still be 'selected' items.

Given the list is pre-populated and you're filtering down rather than searching in my eyes it makes more sense to base it off the MultiSelect.

What do people think?

hakenr:
Do you want to cover a scenario, where there is a huge amount of possible items and you want to display just a few of them (= `HxAutosuggest`-like with `DataProvider`) or do you have a limited number of items you want to provide at once and just add the quick-filter functionality to `HxMultiSelect` (with `Data`)?

In case of `HxMultiSelect`, do you need the input to visualize the filter, or is just "scroll as you type` enough?

bholbrook:
I'd have a limited number of items in the dropdown.
[...]

Issue #453, opened 2023-01-18
[HxMultiSelect] - Ability to have a 'select all' option.


bdsoko:
We would like to have a 'select all' checkbox option and when clicked/checked, would check all available options in the dropdown.  This should be a configurable option to show/hide as needed by the implementation requirements.

PR #617, opened 2023-10-08
[HxMultiSelect] filtering and select all

bholbrook:
# Summary of changes
- Added filtering capability to the HxMultiSelect component as per [Issue 616](https://github.com/havit/Havit.Blazor/issues/616)

- Added select all capability to the HxMultiSelect component as per [Issue 453](https://github.com/havit/Havit.Blazor/issues/453)

Clicking when unchecked will select all items in the currently filtered list.
Clicking when checked will deselect all items in the currently filtered list.
Clicking a checked item or entering any text into the filter will uncheck the select all.
Manually clicking and selecting all items in the filtered list will not check the select all.


Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
hakenr, line 1: Please do not change the order of members in the class. Our coding convention is:
1. parameters (in order of importance)
2. `[Inject]` - dependencies
3. private fields
4. constructor + other methods (usually in order of their appearance in component lifecycle)

Havit.Blazor.Components.Web.Bootstrap/Forms/HxMultiSelect.cs
```diff
@@ -110,6 +110,11 @@ public class HxMultiSelect<TValue, TItem> : HxInputBase<List<TValue>>, IInputWit
 	/// </summary>
 	[Parameter] public RenderFragment InputGroupEndTemplate { get; set; }
 
+	[Parameter] public bool EnableFiltering { get; set; }
```
hakenr, line 113: We don't have any other `EnableXy` parameters for optional features.
This should probably be `Filterable="true"` or `AllowFiltering="true"` or `Filtering="true"`. Not sure now. :-D
(We can change this later before merging to master.)

Havit.Blazor.Components.Web.Bootstrap/Forms/HxMultiSelect.cs
```diff
@@ -202,7 +221,14 @@ private string GetInputText()
 		{
 			return InputText;
 		}
-		else if ((InputTextSelector is null) || (Data is null) || (CurrentValue is null))
+
+		// If filtering is enabled and the dropdown is visible then we want to display the user-entered filter text instead of the selected values summary
+		if (EnableFiltering && IsShown)
+		{
+			return InputText;
+		}
```
hakenr, line 229: (Oops, we misunderstood each other regarding the implementation. I will explain in general comment.)

hakenr (changes requested):
I'm sorry, but I misunderstood the intent, we kind of missed each other. I thought we were talking about a new search field inside the dropdown and I didn't notice that you were describing using an existing input for filtering.
We would actually like to teach the component this UX:
image
image

In our opinion, the filtering functionality combined with the main input is confusing because the selected items are not visible when filtering.
image

I am very sorry, but we cannot accept the filtering functionality in this form. If you would still be willing to give your time to rework it into the intended form, we would be very grateful. Otherwise, we will try to move the request higher in the backlog to appreciate the energy put into the functionality and get the filtering into the component sooner.

Havit.Blazor.Components.Web.Bootstrap/Forms/HxMultiSelect.cs
bholbrook: Thank you for the comments and feedback. I would still like to implement these changes and while I believe the way I've implemented the omni-box to be suitable for my scenario I understand it may not be suitable for everyone.
I shall rework this pull request within the next day or so to meet your standards and suggestions.

BlazorAppTest/Pages/HxMultiSelectTest.razor
```diff
@@ -1,38 +1,61 @@
...
+    }
+
+    private class Model
+    {
+        public List<string> CultureInfos { get; set; }
+    }
+
+    private bool FilterSelector(CultureInfo item, string filter)
+    {
+        return string.IsNullOrWhiteSpace(filter) || item.EnglishName.Contains(filter, StringComparison.OrdinalIgnoreCase);
```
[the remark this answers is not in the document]
bholbrook: I've changed this and the new default FilterPredicate to use string.IsNullOrEmpty.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
bholbrook: Styling has been reverted back to the original.

Havit.Blazor.Components.Web.Bootstrap/Forms/HxMultiSelect.cs
```diff
@@ -110,6 +110,11 @@ public class HxMultiSelect<TValue, TItem> : HxInputBase<List<TValue>>, IInputWit
 	/// </summary>
 	[Parameter] public RenderFragment InputGroupEndTemplate { get; set; }
 
+	[Parameter] public bool EnableFiltering { get; set; }
+
+	[Parameter] public Func<TItem, string, bool> FilterSelector { get; set; } = (_, _) => true;
+
+	private bool IsShown { get; set; }
```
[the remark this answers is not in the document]
bholbrook: A property isn't required. This has been changed to a field.

Havit.Blazor.Components.Web.Bootstrap/Forms/HxMultiSelect.cs
bholbrook: I have reworked this component and believe it's in a good place to re-review.

bholbrook: I've extended this to also include the select all functionality as requested in [Issue 453](https://github.com/havit/Havit.Blazor/issues/453).

hakenr:
@bholbrook Is the PR ready for review now or are you still working on the code?

bholbrook:
@hakenr It's ready to be re-reviewed when you can please :)

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor
```diff
@@ -36,25 +34,60 @@
...
-                    </button>
-                </li>
+                if (AllowSelectAll)
+                {
+                    <li>
+                        <button type="button" class="dropdown-item" role="option" @onclick="HandleSelectAllClickedAsync">
+                            <div class="form-check">
+                                <input class="form-check-input" type="checkbox" id="@($"{InputId}_selectall")" checked="@selectAll" tabindex="-1" />
+                                <label class="form-check-label" for="@($"{InputId}_selectall")" @onclick:preventDefault>
+                                    @(SelectAllText ?? "-- select all --")
```
hakenr, line 64: We have to provide localization for the default text "select all" from the `IStringLocalizer<>`.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -14,6 +16,8 @@ public partial class HxMultiSelectInternal<TValue, TItem>
 
 	[Parameter] public List<int> SelectedIndexes { get; set; }
```
hakenr, line 17: Please remove the parameter, this is dead code (it was no longer being used before the PR).

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
+	[Parameter] public EventCallback<bool> SelectAllChanged { get; set; }
+
+	[Parameter] public string SelectAllText { get; set; }
+
 	/// <summary>
 	/// Additional attributes to be splatted onto an underlying HTML element.
 	/// </summary>
 	[Parameter(CaptureUnmatchedValues = true)] public Dictionary<string, object> AdditionalAttributes { get; set; }
 
+	[Inject] private IJSRuntime JSRuntime { get; set; } = default!;
```
hakenr, line 91: Remove nullable-references notation, we do not use nullable-references feature in our code.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
-	private async Task HandleItemSelectionChangedAsync(bool newChecked, TItem item)
+	public HxMultiSelectInternal()
+	{
+		dotnetObjectReference = DotNetObjectReference.Create(this);
+	}
+
+	/// <inheritdoc cref="ComponentBase.OnAfterRenderAsync(bool)" />
+	protected override async Task OnAfterRenderAsync(bool firstRender)
+	{
+		await base.OnAfterRenderAsync(firstRender);
```
hakenr, line 121: No need to call empty `base` method. We are in a library code where we can be a little bit more aggressive.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
 		{
 			Checked = newChecked,
 			Item = item
 		});
+
+		if (triggerSelectAllEvent)
+		{
+			await ChangeSelectAllAsync(false);
+		}
+	}
```
hakenr, line 152: (I think we should not expose the `SelectAllChanged` event (`EventCallback`) as parameter et al.)
In case we keep the `SelectAllChanged`, please change the method argument name to `resetSelectAll` to better describe the meaning.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
+	private bool DoSelectedValuesContainValue(TValue value)
+	{
+		return SelectedValues?.Contains(value) ?? false;
+	}
+
+	private Task ChangeSelectAllAsync(bool selectAll)
+	{
+		this.selectAll = selectAll;
+		return SelectAllChanged.InvokeAsync(selectAll);
+	}
```
hakenr, line 197: (I think we should not expose the SelectAllChanged event (EventCallback) as parameter et al.)

In case we keep the `SelectAllChanged` parameter, we should check if the `selectAll` value changed and invoke the `EventCallback` only if the value really changed. (Now it gets triggered from `HandleItemSelectionChangedAsync` with every single selection of individual item in the list.)

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
 			Item = item
 		});
+
+		if (triggerSelectAllEvent)
+		{
+			await ChangeSelectAllAsync(false);
+		}
+	}
+
+	private bool selectAll;
```
hakenr, line 154: (Class members order.) Please move the field definition above all the methods where all the other fields are.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
+				if (!itemSelected)
+				{
+					// We don't want to trigger select all triggers each time, an item is selected, just once at the end
+					await HandleItemSelectionChangedAsync(!itemSelected, item, false);
+				}
+			}
+		}
+
+		await ChangeSelectAllAsync(!selectAll);
+	}
```
hakenr, line 186: Not sure we can change the `selectAll` value at the end of this method as it contains many `await` for asynchronnous calls and this might lead to rendering of the component with old `selectAll` value and some other unwanted effects.

Also see overall comment regarding the `ItemSelectionChanged` callback for select-all scenario.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
+		return SelectedValues?.Contains(value) ?? false;
+	}
+
+	private Task ChangeSelectAllAsync(bool selectAll)
+	{
+		this.selectAll = selectAll;
+		return SelectAllChanged.InvokeAsync(selectAll);
+	}
+
+	private Task HandleInputChangedAsync(ChangeEventArgs e)
```
hakenr, line 199: Please rename to `HandleFilterInputChangedAsync' to refer to the filter explicitly.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
```diff
@@ -49,36 +53,252 @@ public partial class HxMultiSelectInternal<TValue, TItem>
...
+	}
+
+	/// <summary>
+	/// Expands the item.
+	/// </summary>
+	public async Task ShowAsync()
+	{
+		await EnsureJsModuleAsync();
+		await jsModule.InvokeVoidAsync("show", elementReference);
+	}
```
hakenr, line 270: Please remove the methods. They are not being used nor exposed to the user.

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxMultiSelect.js
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	element.hxMultiSelectDotnetObjectReference = hxMultiSelectDotnetObjectReference;
+	element.addEventListener('shown.bs.dropdown', handleDropdownShown);
+	element.addEventListener('hidden.bs.dropdown', handleDropdownHidden);
+
+	var d = new bootstrap.Dropdown(element, {
+		toggle: shouldToggle,
+	});
+}
```
hakenr, line 13: Please remove the `shouldToggle` option, it is always `false`.

hakenr (commented):
@bholbrook First off, I have to commend the effort you've put into working on changes of this scale. I myself was surprised by the sheer amount of changes, which is partly due to the combination with adding the "select all" feature. This raises a lot of new questions, although I understand that implementing both features together is more efficient.

For now, I've added some inline code comments and have noted the following points (for myself) that need to be addressed. Most of these will require further thought on my part, and some will be discussed with colleagues in our internal design meeting:

- [ ] `AllowFiltering`, `AllowSelectAll` naming?
- [x] `EmptyTemplate` naming. Perhaps `FilterEmptyResultTemplate` and `FilterEmptyResultText` (with defaults) for easier usage?
- [ ] SelectAll when filtered - should it select only visible items? Is that what we want?
- [ ] Filter-box - should it have a search icon and a clear icon?
- [x] `ClearFilterOnHide` - should the default be `true`?
- [x] `OnShown` and `OnHidden` - do we even need these parameters? How would anyone use them?
- [x] `MultiSelectSettings` - `AllowFiltering`, `AllowSelectAll`, `ClearFilterOnHide`, ?`FilterPredicate`?
- [ ] `SelectAllChanged` - do we even need such an `EventCallback`? How would it be used? If so, maybe `OnSelectAllChanged`, and it should not have a value.
- [ ] Missing documentation for new parameters
- [ ] Missing demo documentation for new features
- [ ] Default "-- select all --" text - do we want those dashes? If so, probably just one on each side?
- [ ] `ItemSelectionChanged` - we probably can't use this for the SelectAll scenario, as it would trigger a series of `ValueChanged` for each selected item, whereas it might be more appropriate to trigger `ValueChanged` just once with a `List<TValue>` of all items
- [ ] `InputSize` to be applied to inner filter-input

Looking forward to your thoughts on these points.

Havit.Blazor.Components.Web.Bootstrap/Forms/HxMultiSelect.cs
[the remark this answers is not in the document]
hakenr: Please rename to `isDropdownShown` to better refer to the dropdown part of the component.

hakenr:
@crdo, this is a significant contribution that we shouldn't let sit idle. Could you please take a look at the PR (UX, HTML, API, ...) when you get a chance? Thanks!

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor
```diff
@@ -36,25 +34,60 @@
     <ul class="dropdown-menu"> @* Must be always rendered otherwise does not work after disable->enabled scenario *@
         @if (enabled)
         {
-            for (int i = 0; i < ItemsToRender.Count; i++)
+            if (AllowFiltering)
+            {
+                <input @ref="filterInputReference"
+                       id="@($"{InputId}_filter")"
+                       type="text"
+                       class="@CssClassHelper.Combine("form-control")"
```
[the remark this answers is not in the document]
bholbrook: Have removed the redundant call.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor
bholbrook: I've reworked this to call a helper method which uses the newly injected IStringLocalizer with new resource file. I've only added English for now as that's all I'm comfortable with.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
bholbrook: I have removed this property and the Values property which was also not in use.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor.cs
bholbrook: Have removed these methods.

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxMultiSelect.js
bholbrook: Have removed the toggle option

bholbrook:
@hakenr In response to your list, these are my thoughts.

* [ ]  `AllowFiltering`, `AllowSelectAll` naming?
I don't mind what these are called as long as it's obvious they are a boolean on/off feature using an action verb prefix.
Happy to change them to whatever you want, but I'd rather have the functionality committed than a perfected name.

* [ ]  `EmptyTemplate` naming. Perhaps `FilterEmptyResultTemplate` and `FilterEmptyResultText` (with defaults) for easier usage?
Have renamed the existing template parameter to FilterEmptyResultTemplate.
Have added a new FilterEmptyResultText parameter.
Have added a default empty template utilising this new FilterEmptyResultText parameter. This also uses localisation.

* [ ]  SelectAll when filtered - should it select only visible items? Is that what we want?
I definitely think so. You have to actively filter and you have to have to actively click the select all button, which means when you do click select all you've actively chosen to select what you see contextually.
I can't think of a single real-world scenario when you'd want to select every item in the list while you've got a filter.

* [ ]  Filter-box - should it have a search icon and a clear icon?
A search icon isn't required as it will update the list immediately as the user types.
A clear icon has now been added and has been added to the settings too.
[IN_PROGRESS] Icon should be placed inside the filter input box.

* [ ]  `ClearFilterOnHide` - should the default be `true`?
I have changed the default to true.
As a note: When the dropdown is expanded the filter is automatically highlighted so the user can just hit delete so even if they do actively set it to false the filter is still very usable.

* [ ]  `OnShown` and `OnHidden` - do we even need these parameters? How would anyone use them?
I have removed these.

* [ ]  `MultiSelectSettings` - `AllowFiltering`, `AllowSelectAll`, `ClearFilterOnHide`, ?`FilterPredicate`?
Added AllowFiltering and AllowSelectAll to the MultiSelectSettings and set both defaults to false.
Added ClearFilterOnHide to the MultiSelectSettings and set the default to true.

* [ ]  `SelectAllChanged` - do we even need such an `EventCallback`? How would it be used? If so, maybe `OnSelectAllChanged`, and it should not have a value.
I have removed this.

* [ ]  Missing documentation for new parameters
All documentation has been added to `HxMultiSelect` and removed from `HxMultiSelectInternal`

* [ ]  Missing demo documentation for new features
I have added new examples for basic filtering, custom filtering, and a simple select all example.
I don't know how to run the documentation project locally so haven't been able to test these :(
How can I run the documentation to check please?

* [ ]  Default "-- select all --" text - do we want those dashes? If so, probably just one on each side?
I've changed this to default to "-select all-".

* [ ]  `ItemSelectionChanged` - we probably can't use this for the SelectAll scenario, as it would trigger a series of `ValueChanged` for each selected item, whereas it might be more appropriate to trigger `ValueChanged` just once with a `List<TValue>` of all items

hakenr:
@bholbrook

### SelectAll when filtered - should it select only visible items? Is that what we want?
>I definitely think so. You have to actively filter and you have to have to actively click the select all button, which means when you do click select all you've actively chosen to select what you see contextually.
> I can't think of a single real-world scenario when you'd want to select every item in the list while you've got a filter.

I agree that you don't want to "select all including those filtered out", but we should somehow express this in the UX (text?) when filtering is applied. Maybe something like "add filtered items to selection" (Huh, do we want to add the items to existing selection or do we want to replace the selection?).

### Filter-box - should it have a search icon and a clear icon? [IN PROGRESS]
> A search icon isn't required as it will update the list immediately as the user types.

This was meant as a visual indicator "this is the input for filtering" not as "spinner" while loading the data.
image

###  SelectAllChanged - do we even need such an EventCallback? How would it be used? If so, maybe OnSelectAllChanged, and it should not have a value.
> I have removed this.

Have you pushed all the changes? It is still there with all the related code.

### ItemSelectionChanged - we probably can't use this for the SelectAll scenario, as it would trigger a series of ValueChanged for each selected item, whereas it might be more appropriate to trigger ValueChanged just once with a List<TValue> of all items
>TBD

We probably have to change `SelectionChangedArgs` to allow multiple changes to be signaled at once. Probably something like
```csharp
public class SelectionChangedArgs
{
	public List<TItem> ItemsDeselected { get; set; }
	public List<TItem> ItemsSelected { get; set; }
}
```
The `HandleItemSelectionChanged` can then change the `CurrentValue` accordingly.

bholbrook:
@hakenr Haha, apparently I made the SelectAllChanged changes in my head! I've pushed them up now.

In terms of the SelectAll filtering language, I'm really not sure what to do here. If you have any suggestions I'm happy to make those changes.

I'm currently working on the Search/Clear icon. I didn't about having the Search icon there but that's not a bad idea.
I'll attempt to implement it similarly to the others whereby the Search icon will display while the filter box is empty and the Clear icon will display when there is any text.

And the ItemSelectionChanged stuff I haven't yet investigated. It's likely the last thing I'll work on.

bholbrook:
@hakenr I've just pushed up the final changes to your list of suggestions.
With them, everything should be ready to go.
If there is anything you think I should change (outside of documentation because I don't know how to run that project) please do let me know and I'll gladly make those changes.
I hope to get this integrated soon!

Also, once this does go in, how long might it be before this gets deployed to NuGet and made available?

crdo (changes requested):

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor
```diff
@@ -36,25 +34,67 @@
     <ul class="dropdown-menu"> @* Must be always rendered otherwise does not work after disable->enabled scenario *@
         @if (enabled)
         {
-            for (int i = 0; i < ItemsToRender.Count; i++)
+            if (AllowFiltering)
+            {
+                <input @ref="filterInputReference"
```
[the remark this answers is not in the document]
bholbrook: I've wrapped the filter input and empty filter templates in `<li>` and have added a 0.25em padding in the filter input.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxMultiSelectInternal.razor
```diff
@@ -36,25 +34,67 @@
...
+                       id="@($"{InputId}_filter")"
+                       type="text"
+                       class="form-control"
```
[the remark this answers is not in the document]
bholbrook: I have updated this such that the filter box now respects the parent input size.

hakenr (approved):
LGTM.
Will do some minor tuning and will merge to master.

PR #626, opened 2023-10-16
fixed #621 [HxInputDate] Add CalendarDisplayMonth

### LINKED ISSUE #621: [HxInputDate] Add DisplayMonth to HxInputDate
Hello,

I have a requirement to add the `HxCalendar.DisplayMonth` parameter to the `HxInputDate`.

More than happy to implement and raise a PR.

That is quite a simple change.

Part 2 is for `HxInputDateRange`, it would be a bit more complex as the whole idea is to optionally change the `DisplayMonth` of the `EndDate` when the `StartDate` is changed while `EndDate` is still null.  Should it be a flag along the lines of end date is after start date?  Think holiday reservation or event booking.

Any thoughts before I start implementation?

Thanks.

blanchardglen:
[HxInputDateRange] Bug ToInputDateRange displays the From selected date
Add FromCalendarDisplayMonth and ToCalendarDisplayMonth
The To calendar will choose the latest of the FromDate or ToCalendarDisplayMonth or if both are not set then the FromCalendarDisplayMonth

Any changes required please let me know.

### DIFF
--- a/BlazorAppTest/Pages/HxInputDate_Issue621_Test.razor
+++ b/BlazorAppTest/Pages/HxInputDate_Issue621_Test.razor
@@ -6,10 +6,10 @@
 <HxInputDate Label="No DisplayMonth" @bind-Value="value1"/>
 <HxInputDate Label="Next Month Value" @bind-Value="nextMonth" />
 <HxInputDate Label="Next Month Display Month" @bind-Value="value2" CalendarDisplayMonth="@nextMonth" />
-<HxInputDateRange Label="DateRange No DisplayMonth" @bind-Value="Holidays" />
-<HxInputDateRange Label="DateRange From DisplayMonth" @bind-Value="Unset" FromCalendarDisplayMonth="@nextMonthNextYear" />
-<HxInputDateRange Label="DateRange To DisplayMonth" @bind-Value="Unset" ToCalendarDisplayMonth="@nextMonthNextYear" />
-<HxInputDateRange Label="DateRange Both DisplayMonth" @bind-Value="Unset" FromCalendarDisplayMonth="@nextMonth" ToCalendarDisplayMonth="@nextMonthNextYear"/>
+<HxInputDateRange Label="DateRange No DisplayMonth" @bind-Value="holidays" />
+<HxInputDateRange Label="DateRange From DisplayMonth" @bind-Value="unset1" FromCalendarDisplayMonth="@nextMonthNextYear" />
+<HxInputDateRange Label="DateRange To DisplayMonth" @bind-Value="unset2" ToCalendarDisplayMonth="@nextMonthNextYear" />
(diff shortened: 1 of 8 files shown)


BlazorAppTest/Pages/HxInputDate_Issue621_Test.razor
```diff
@@ -0,0 +1,33 @@
...
+<HxInputDateRange Label="DateRange Both DisplayMonth" @bind-Value="Unset" FromCalendarDisplayMonth="@NextMonth" ToCalendarDisplayMonth="@NextMonthNextYear"/>
+
+@code {
+    private DateTime NextMonth;
+    private DateTime NextMonthNextYear;
+    private DateTime Value1;
+    private DateTime Value2;
+
+    private DateTimeRange Holidays;
+    private DateTimeRange Unset;
```
hakenr, line 21: `camelCase` for private fields.

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateRangeInternal.razor.cs
```diff
@@ -47,6 +50,27 @@ public partial class HxInputDateRangeInternal : InputBase<DateTimeRange>, IAsync
 	private HxDropdownToggleElement fromDropdownToggleElement;
 	private HxDropdownToggleElement toDropdownToggleElement;
 
+	private DateTime toCalendarDisplayMonthEffective
```
hakenr, line 53: Please make this a method called `GetToCalendarDisplayMonthEffective()`

hakenr (changes requested):
I think, if there is a `Value`, it should be always used for displayed month. The new parameters should be effective only when there is no `Value` present. (Separate behavior for `From`/`To` in `HxInputDateRange`).

blanchardglen:
> I think, if there is a `Value`, it should be always used for displayed month. The new parameters should be effective only when there is no `Value` present. (Separate behavior for `From`/`To` in `HxInputDateRange`).

The HxCalendar takes care of this already in my testing, please correct me if I am wrong though.

The only minor complexity is in the `GetToCalendarDisplayMonthEffective` for `HxInputDateRange`  because the `From` value might override the `ToCalendarDisplayMonth` but only if it is less than the provided `ToCalendarDisplayMonth`

I would imagine the vast majority of times the `ToCalendarDisplayMonth` would never be set, in that case should it be removed?

hakenr:
Current behavior to fix (calendar resets to `DisplayMonth` although there is `Value` from previous month):

hakenr:
> I would imagine the vast majority of times the `ToCalendarDisplayMonth` would never be set, in that case should it be removed?

As you already implemented it, if the `Value` will have precedence, let's keep the `ToCalendarDisplayMonth` for those rare cases when you expect the range to end in some specific month in future 🤷‍♂️.

blanchardglen:
Thanks for your patience!

Adjusted for `Value` to take precedence in `HxInputDate` and both `From` and `To` parts of `HxInputDateRange`

hakenr (approved):
lgtm

PR #667, opened 2023-11-19
HxSidebar Toggler v2 WIP

crdo:
@hakenr I have implemented a new `HxSidebar` `Toggler` inspired by chat.openai.com.
It is little more minimalistic and shows a tooltip on hover.

What needs to be done
- [x] Get rid of ExpandIcon and CollapseIcon paramters
- [x] Add resources for Expand/Collapse text in tooltip
- [x] Cleanup the CSS with some variables so it is a bit more cleaner and configurable

I would like to make it into next major release.


hakenr (approved):
lgtm

Issue #687, opened 2023-12-01
[HxSidebar] highlight parent as active when its child is active


alexax578:
I want to be able to configure HxSidebar in a way that makes it's parent sidebar items highlighted as active when any of their children are active:
![image](https://github.com/havit/Havit.Blazor/assets/37944469/43717023-8894-4187-b78d-1e7078e88998)

PR #688, opened 2023-12-01
#687 New parameter HighlightOnActiveChild + CSS

crdo:


Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor
```diff
@@ -1,6 +1,6 @@
 ﻿@namespace Havit.Blazor.Components.Web.Bootstrap
 <CascadingValue Value="this" IsFixed="true">
-    <div>
+    <div class="@CssClassHelper.Combine(HighlightOnActiveChild ? "highlight-on-active-child" : null)">
```
hakenr, line 3: The name of the class should start with `.hx-sidebar-item-...`

Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor.cs
```diff
@@ -36,6 +36,12 @@ public partial class HxSidebarItem : IAsyncDisposable
 	/// </summary>
 	[Parameter] public bool ExpandOnMatch { get; set; } = true;
 
+	/// <summary>
+	/// Set to <c>false</c> if you don't want to highlight the item if one of the children URL matches.<br/>
+	/// Default is <c>true</c>.
+	/// </summary>
+	[Parameter] public bool HighlightOnActiveChild { get; set; } = true;
```
hakenr, line 43: In current state, I would probably go for `false` as the default.
Most of the sidebars I found are not highlighting the parent node et al.
A few do, but usually in different (less accented) style that the item itself.
The visual I like is GitHub:
image

...consider making the visual style separately defined.
...consider "just bold" for the default parent highlighting (in such case I would vote for `true` as default value :-D)

crdo: All done.

hakenr (approved):
lgtm

Issue #686, opened 2023-12-01
[HxSidebar] Add support for tooltips on items & display dropdown on hover when collapsed


alexax578:
I want to be able to set a tooltip text to individual sidebar items.

- The tooltip should be visible only when I hover on the item when the sidebar is collapsed.
- The tooltip can stay visible when I open the submenu with child navigation items but should appear only when I hover over the parent icon.

In my scenario, tooltips are only relevant on the first-level parent sidebar items.

hakenr:
(During the implementation, the scope was widened to include "dropdown on hover".)

Issue #204, opened 2022-06-07
[CONCEPT] Allow combining CSS-classes within component customization - `CoreCssClass`


hakenr:
## Problem description
Currently, when you set
```csharp
HxSomething.Defaults.XyCssClass = "default-class";
```
and then set the class in markup code
```razor
<HxSomething XyCssClass="local-class" />
```
the `default-class` is lost and overridden by `local-class` assignment.

For better customizability (especially application-wide settings) we should be able to combine CSS classes from settings with instance CSS classes (local markup etc.) to get both classes rendered `default-class local-class`.

## Solution proposal
Some component libraries support this through additional syntax in CssClass-properties, such as `+default-class`, but this is probably not the way we would like to go.

Some of our component already use `CoreCssClass` properties/methods which behave this way (they get added to the `CssClass` in the end).

I think we should consolidate the current definitions and usages of `[Xy]CoreCssClass` and add the ability to set them through `Settings` and `Defaults`. This should solve the issue we are facing...

@crdo @jirikanda What do you think?

PR #689, opened 2023-12-01
#686 [HxSidebar] Add support for tooltips on items WIP

crdo:
#686

### DIFF
--- a/Havit.Blazor.Components.Web.Bootstrap/HxSetup.cs
+++ b/Havit.Blazor.Components.Web.Bootstrap/HxSetup.cs
@@ -16,7 +16,10 @@ public static class HxSetup
 	/// </remarks>
 	public static string RenderBootstrapJavaScriptReference()
 	{
-		return "<script src=\"https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/js/bootstrap.bundle.min.js\" integrity=\"sha384-C6RzsynM9kWDrMNeT87bh95OGNyZPhcTNXj1NW7RuBCsyN/o0jlpcV8Qyq46cDfL\" crossorigin=\"anonymous\"></script>";
+		return """
+		       <script src="https://cdn.jsdelivr.net/npm/@popperjs/core@2.11.8/dist/umd/popper.min.js" integrity="sha384-I7E8VVD/ismYTF4hNIPjVp/Zjvgyol6VFvRkX/vR+Vc4jQkC+hVqc2pM8ODewa9r" crossorigin="anonymous"></script>
+		       <script src="https://cdn.jsdelivr.net/npm/bootstrap@5.3.2/dist/js/bootstrap.min.js" integrity="sha384-BBtl+eGJRgqQAUMxJ7pMwbEyER4l1g+O15P+16Ep7Q9Q+zqX6gSbd85u4mG4QzX+" crossorigin="anonymous"></script>
+		       """;
 	}
 
 	/// <summary>
--- a/Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor.cs
+++ b/Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor.cs
@@ -57,6 +57,11 @@ public partial class HxSidebarItem : IAsyncDisposable
 	/// </summary>
 	[Parameter] public RenderFragment ChildContent { get; set; }
 
+	/// <summary>
+	/// Text content of tooltip displayed while hovering item of collapsed Sidebar.
+	/// </summary>
+	[Parameter] public string TooltipText { get; set; }
+
 	[CascadingParameter] protected HxSidebar ParentSidebar { get; set; }
 	[CascadingParameter] protected HxSidebarItem ParentSidebarItem { get; set; }
 	[CascadingParameter] protected HxDropdown DropdownContainer { get; set; }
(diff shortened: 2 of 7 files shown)


crdo:
@alexax578 

1. I need to initialize Bootstrap tooltip in HxSidebar as per https://getbootstrap.com/docs/5.3/components/tooltips/#enable-tooltips
2. The data-bs-toggle="@((!this.expanded && ParentSidebar.Collapsed) ? "tooltip" : null)" on line 3 in HxSidebarItem.razor is not working as expected

Havit.Blazor.Components.Web.Bootstrap/HxSetup.cs
hakenr, line 1: Why do we want to link the Popper separately? Do we need a newer version than the one bundled with Bootstrap?
...does not make sense to me, quite an aggressive approach.

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxSidebarItem.js
```diff
@@ -0,0 +1,75 @@
...
+	dropdown.popperInstance.destroy();
+
+	// Remove the event listeners
+	dropdown.navElement.removeEventListener('mouseenter', dropdown.show);
+	dropdown.navElement.removeEventListener('focus', dropdown.show);
+
+	dropdown.navElement.removeEventListener('mouseleave', dropdown.hide);
+	dropdown.navElement.removeEventListener('blur', dropdown.hide);
+
+	dropdowns.delete(dropdownId);
```
hakenr, line 74: ...all the code has to be guarded with `if (dropdown)`, otherwise we will hit usual trouble where the DOM is manipulated earlier that the destroy/dispose logic runs.

Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor.cs
```diff
@@ -99,13 +108,35 @@ protected override void OnParametersSet()
...
 	protected bool HasExpandableContent => (this.ChildContent is not null);
+	protected bool ShouldInitializePopper => HasExpandableContent && ParentSidebar.Collapsed;
 
 	protected override async Task OnAfterRenderAsync(bool firstRender)
 	{
+		if (firstRender)
+		{
+			await EnsureJsModule();
+		}
+
```
hakenr, line 119: Probably all the JS-related stuff should be conditional under `ShouldInitializePopper`?

Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor.cs
```diff
@@ -133,6 +164,12 @@ protected virtual async ValueTask DisposeAsyncCore()
 		{
 			await ParentSidebarItem.ChildItemsRegistration.UnregisterAsync(this);
 		}
+
+		if (popperInitialized)
+		{
+			popperInitialized = false;
+			await jsModule.InvokeVoidAsync("destroyPopper", navId, id);
+		}
```
hakenr, line 172: Missing `jsModule.Dispose` + usual `try-catch` logic to handle Blazor Server scenarios (circuit closed before disposal logic finishes etc.)

hakenr (changes requested):

Havit.Blazor.Components.Web.Bootstrap/Navigation/HxSidebarItem.razor.cs
alexax578: The `destroyPopper` method is meant to represent `dispose`. Should I rename it?

Havit.Blazor.Components.Web.Bootstrap/wwwroot/HxSidebarItem.js
alexax578: Actually when this code is called the nav element is not present in the DOM. That is why I capture a reference to the element in the `dropdowns` variable and use this reference to cleanup the event listeners. Since I capture this reference to the element I believe I don't have to guard it with an `if (dropdown)` clause.

On the other hand I can also leave it up to the browser -> modern browsers clean up event listeners when the element they are attached to is removed from DOM (when no other references to the element exist).

Havit.Blazor.Components.Web.Bootstrap/HxSetup.cs
alexax578: If I understand the Bootstrap bundle correctly, it does not export the Popper dependency and therefore I am unable to use it.

alexax578:
@crdo I couldn't find a smarter way to hide/show tooltip other than an if/else branch that renders the whole structure wrapped in a tooltip and without a tooltip component.

Havit.Blazor.Components.Web.Bootstrap/HxSetup.cs
hakenr: 😭 Unfortunately, for projects not using the `HxSetup.RenderBootstrapJavaScriptReference()`, this is a major breaking change.

Given that many projects likely fall into this category (since the use of a server-rendered `.cshtml` host page wasn't initially included in the Visual Studio Blazor templates and the `HxSetup` methods were added as optional), we can't implement this change in a minor version of the library.

If we won't be able to find a way without referencing "standalone Popper", we'll need to schedule this for the next major release (version 5). However, please be aware that assembling this release will take some time, as we plan to incorporate several other fundamental changes, like the `CoreCssClass` feature in issue #204."

(We can still consider creating a pre-release of v5 which you might use in the project, but...)

I will now spend some time trying to find a way to achieve the result without using Popper directly.

hakenr:
@crdo @alexax578 I'm proposing a 'competitive' implementation in #703 where the same effect is achieved without direct usage of PopperJS.
(The branch currently does not include the tooltips. We can add them if we agree on this solution.)

hakenr:
abandoned, implemented in #703

PR #703, opened 2023-12-14
#686 [HxSidebar] Add support for tooltips on items & display dropdown…

hakenr:
… on hover when collapsed - part 1 - display on hover (using Bootstrap, no direct access to Popper needed)


hakenr:
(@crdo As I found the cause in [0,2] ofset, I reverted all the CSS changes. If there is anything useful from the original PR, feel free to fine-tune the CSS.)

hakenr:
TBD:
- [ ] Do we want the `TooltipText` to use a default value of `Text` parameter?
- [ ] Do wa want to display the tooltip for sub-items?
cc @crdo @alexax578

PR #724, opened 2024-01-05
[HxCheckbox] Checkbox inside input-group is misaligned

### LINKED ISSUE #460: [HxCheckbox] Checkbox inside input-group is misaligned
When we render `HxCheckbox` inside `input-group-text`, it has an extra `form-check` class that makes it misaligned.

With extra class
image

Without
image

alexax578:
Resolves #460.


crdo:
@hakenr judging by this https://getbootstrap.com/docs/5.3/forms/checks-radios/#without-labels it feels right.

Issue #719, opened 2024-01-02
[HxSidebar] HighlightOnActiveChild styles are not applied when Sidebar is collapsed


alexax578:
This selector `.hx-sidebar-item-highlight-on-active-child:has(.hx-sidebar-subitems .nav-link.active) ::deep > a.nav-link` that applies styles to the parent when a child is active does not work when the sidebar is collapsed because of this wrapping div that is added:
![image](https://github.com/havit/Havit.Blazor/assets/37944469/e8b2945a-9568-44e0-a853-f6c77a3efef3)

alexax578:
I have an issue where the selector targets icon of the parent item as well as icons of all the children items.

crdo:
@alexax578 could you please create a reproduction page in TestApp?

alexax578:
@crdo I adjusted the current demo of HxSidebar in the test app to demonstrate this behavior when you click on a child by setting a different color of --hx-sidebar-parent-item-active-icon-color. If it is desired I can create a separate reproduction page.

PR #740, opened 2024-01-24
[HxSidebar] HighlightOnActiveChild styles are not applied when Sidebar is collapsed

alexax578:
Resolves issue with the solution of #719 where the selector would target all childrens icons as well as the parents icon.


crdo (approved):

hakenr (approved):
lgtm (checked with @crdo)

PR #743, opened 2024-01-25
[HxSwitch] Classes form-check form-switch missing

alexax578:
I totally missed the scenario of HxSwitch that is built upon the logic of **HxCheckbox**. I have adjusted the logic as follows:

- Created a separate property for a CSS class named **AdditionalFormElementCssClass** which is always applied and allows **HxSwitch** to apply it's _form-switch_ class.
- Adjusted the structure of the private properties and logic of when and where the _form-check_ CSS class is applied.
- When no Text is present the _form-check_ CSS class should be omitted according to https://getbootstrap.com/docs/5.3/forms/checks-radios/#without-labels


hakenr (approved):
lgtm

Issue #745, opened 2024-01-26
Extend ThemeColorExtensions class with ToBorderColorCss and subtle/emphasis variants


crdo:
My motivation is to get an API of our a components from 
```
<ReportAggregateValueWidget Icon="BootstrapIcon.DatabaseDown"
			BackgroundColor="@(_selectedSeriesIds.Contains((int)InvoicesReportSettingsModel.ReportValue.Costs) ? "rgba(var(--bs-danger-rgb), 0.05)" : null)"
			IconBackground="rgba(var(--bs-danger-rgb), 0.2)"
			IconColor="rgb(var(--bs-danger-rgb))" />
```

to
```
<ReportAggregateValueWidget Icon="BootstrapIcon.DatabaseDown"
			Color="ThemColor.Danger"
			Active="@(_selectedSeriesIds.Contains((int)InvoicesReportSettingsModel.ReportValue.Costs)" />
```
and then do the design work inside of the component with ThemeColorExtensions.

PR #746, opened 2024-01-26
#745 Added ToBorderColorCss + emphasis/subtle variants

crdo:
In this PR I added ToBorderColorCss extension method to be able to get border css class based on the ThemeColor and also added subtle/emphasis variants to ToBackgroundColorCss and ToTextColorCss.

Also created TestPage at BlazorAppTest/ThemeColorExtensionsTest.


hakenr (approved):
lgtm

I'm still unsure whether `ThemeColor` should remain an `enum` or if we'll need to implement some extensibility there. However, let's leave it as is for now and hopefully, it won't complicate things later on.

PR #772, opened 2024-03-20
Add search icon placement to HxSearchBox component

crdo:
In this PR, i have added a SearchIconPlacement enum and adjusted the HxSearchBox to be able to set the search icon positiont o left or right.

I have also removed a condition on input where right padding was set only if cear icon was visible. The padding should be set always.


hakenr (approved):
lgtm

PR #792, opened 2024-04-17
Bump NuGet packages. Fix versions for .NET 6 and 7.

EnchantedCoder:
I noticed that NuGet package restore would fail for .NET 6 and 7 due to some package references not having framework targets set, in addition to dependencies being out of date. This change fixes that.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -6,41 +6,47 @@
 	<ItemGroup>
 		<PackageVersion Include="Blazored.FluentValidation" Version="2.1.0" />
 		<PackageVersion Include="bunit.web" Version="1.27.17" />
-		<PackageVersion Include="Grpc.AspNetCore.Web" Version="2.61.0" />
-		<PackageVersion Include="Grpc.Net.Client" Version="2.61.0" />
-		<PackageVersion Include="Grpc.Net.Client.Web" Version="2.61.0" />
+		<PackageVersion Include="Grpc.AspNetCore.Web" Version="2.62.0" />
+		<PackageVersion Include="Grpc.Net.Client" Version="2.62.0" />
+		<PackageVersion Include="Grpc.Net.Client.Web" Version="2.62.0" />
 		<PackageVersion Include="Havit.Core" Version="2.0.29" />
 		<PackageVersion Include="Havit.AspNetCore" Version="2.0.12" />
-		<PackageVersion Include="LoxSmoke.DocXml" Version="3.6.0" />
-		<PackageVersion Include="Microsoft.AspNetCore.Components.Web" Version="8.0.3" Condition="'$(TargetFramework)' == 'net8.0'" />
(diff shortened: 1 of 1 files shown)


Directory.Packages.props
```diff
@@ -6,41 +6,47 @@
...
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly" Version="7.0.18" Condition="'$(TargetFramework)' == 'net7.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly" Version="6.0.29" Condition="'$(TargetFramework)' == 'net6.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="8.0.4" Condition="'$(TargetFramework)' == 'net8.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="7.0.18" Condition="'$(TargetFramework)' == 'net7.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.DevServer" Version="6.0.29" Condition="'$(TargetFramework)' == 'net6.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.Server" Version="8.0.4" Condition="'$(TargetFramework)' == 'net8.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.Server" Version="7.0.18" Condition="'$(TargetFramework)' == 'net7.0'" />
+		<PackageVersion Include="Microsoft.AspNetCore.Components.WebAssembly.Server" Version="6.0.29" Condition="'$(TargetFramework)' == 'net6.0'" />
```
hakenr, line 29: @EnchantedCoder What type of NuGet package restore would failure did you encounter?
These packages (3) are only used in the documentation, which is `net8.0`-only and does not use multi-targeting.

hakenr (commented):
@EnchantedCoder Thanks for the contribution and for updating the packages. Could you please clarify what type of issues you were addressing? I haven't encountered any problems yet, but the central package management can still be quite bumpy, especially with tooling.

Directory.Packages.props
EnchantedCoder: Havit.Blazor.Grpc.Client was failing to restore packages, complaining that the 8.0.x version is only compatible with .NET 8, as soon as I updated to 8.0.4. Making that change in the central package management file allowed it to restore and build normally again.

hakenr: Interesting, I will try to replay that scenario on my side.
(`Havit.Blazor.Grpc.Client` does not reference any of the `Microsoft.AspNetCore.Components.WebAssembly.*` packages.)

hakenr: Merged into master, tried to remove the .NET6 + 7 PackageVersions for those three packages and still able to restore (incl. `dotnet restore -f`).

@EnchantedCoder I will push that commit to master. Can you please try to double-check on your side?

EnchantedCoder:
Strange. Visual Studio 2022 must have just been confused, then.

PR #797, opened 2024-04-26
ComponentApiDocModelBuilder - fix enum summary order

Harvey1214:
The summaries for the enum members weren't being retrieved according to a reliable key - the correct summaries were matched only if the enum values started from 0 and inscreased by one. In most cases it resulted in the following.

image


hakenr (approved):
lgtm

PR #836, opened 2024-06-21
Unify clear button color of HxSearch and HxAutosuggest + add possibility to set icon colors via CSS vars

crdo:
Unified the look of clear `color` in `HxSearch` and `HxAutosuggest`.
Added CSS var for setting `color` of search and clear icon with unset as default.
Updated docs.


hakenr (approved):
lgtm

Issue #118, opened 2022-04-19
[HxGrid] Context menu causes scrollbar


Apskaita5:
When there are few rows in the grid and a relatively large context menu, when showing context menu it doesn't fit the table area which forces a scrollbar to appear (part of context menu hidden).

hakenr:
@crdo, can you please take a look and asses this issue? Thanx.

crdo:
@Apskaita5 @hakenr yes, I have came across this issue already. @Harvey1214 please create a reproducing page for this issue in our BlazorTestApp so I can try to come up with some solution.

Harvey1214:
@crdo the reproducing page is completed (https://dev.azure.com/havit/DEV/_git/002.HFW-HavitBlazor/commit/6ade98d68b9ffa9a1447636901ebc2d0f26fdf00).

crdo:
Anyway, what causes your issue most likely is the presence of `overflow: auto | hidden;` on some of the parent elements.
In this case you have two options:

1. Get rid of the overflow
2. Tweak the Popper config so the Dropdown (that is used internally for the HxContextMenu) behaves as close as desired https://popper.js.org/docs/v2/modifiers/prevent-overflow/

@hakenr can we set our own Popper config to the relevant HxCompoments?

Apskaita5:
> @Apskaita5 could you please confirm that the HxGrid instance that you refer to is responsive (IsResponsive=true)?

Yep. (CSS table-responsive)

Apskaita5:
> Anyway, what causes your issue most likely is the presence of `overflow: auto | hidden;` on some of the parent elements. 

Only if it is present on Havit components.

crdo:
Related discussion in Bootstrap repo https://github.com/twbs/bootstrap/issues/33463#issuecomment-822253936

crdo:
> As far as I understand, new Bootstrap release should resolve this issue with https://github.com/twbs/bootstrap/pull/33684 and we don't need to do anything on our side. @crdo?

I think that this issue has been already merged into one of the previous versions.

hakenr:
Already reported to Bootstrap: https://github.com/twbs/bootstrap/issues/35774

hakenr:
Bootstrap 5.2.0 probably does not offer any suitable solution (setting boundary with data-attribute or something like that).

It is probably achievable with JavaScript options (setting boundary to body): https://github.com/twbs/bootstrap/issues/36358
[...]

PR #844, opened 2024-06-27
Attribute splatting for HxContextMenu dropdown trigger element

crdo:
Adressing Issue #118 [HxGrid] Context menu causes scrollbar

### DIFF
--- a/BlazorAppTest/Pages/HxGrid_Issue118_Test.razor
+++ b/BlazorAppTest/Pages/HxGrid_Issue118_Test.razor
@@ -10,7 +10,7 @@
         <HxGridColumn TItem="CultureInfo" HeaderText="Name" ItemTextSelector="@(item => item.Name)" SortKeySelector="@(item => item.Name)" />
         <HxGridColumn TItem="CultureInfo" HeaderText="EnglishName" ItemTextSelector="@(item => item.EnglishName)" SortKeySelector="@(item => item.EnglishName)" />
         <HxContextMenuGridColumn Context="item">
-            <HxContextMenu>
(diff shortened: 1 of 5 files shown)


hakenr (commented):
Although I understand the current need (#118), I'm not satisfied with the selective decision we are making here when splatting the additional attributes to *some* inner `.hx-context-menu-btn` element (`div.hx-context-menu > div.dropdown > div.hx-context-menu-btn`).

In the case of `HxInputBase` components, the additional attributes are splatted to the `input` (to maintain the `InputBase` approach). All the other components (hopefully) are splatting the additional attributes to the dominant element (the outermost one).

We should find another way to achieve the goal. Is there any relevant scenario where the `data-bs-popper-config='{"strategy":"fixed"}'` should not be used? Can we set this config always?

crdo:
I don't think it is a good idea to make `"strategy":"fixed"` default. If the `.dropdown-menu` element is set as `position: fixed;` its position is calculated and updated when you scroll the window. While with `position: absolute;`, it is calculated on the first render and thats it.

crdo:
On the other hand - if you start scrolling while the `HxContextMenu` is open, it disappears as you no longer hover the row.

hakenr:
New `PopperStrategy="[Absolute|Fixed]"` parameter?

crdo:
Either PopperStrategy or PositionStrategy works for me.

hakenr:
Added new `PopperStrategy` parameter to `HxContextMenu`, which allows to fix the issue:
```razor
  <HxContextMenu PopperStrategy="PopperStrategy.Fixed">
```
If you feel this should be configurable as application-wide default with `HxContextMenu.Defaults.PopperStrategy`, please create a separate issue.

Issue #604, opened 2023-09-15
[HxTreeView] Collapse area smaller than expand


hakenr:
Left of the carret, there is a narrow area which allows me to expand the node, but does not collapse the node when clicked again:
![image](https://github.com/havit/Havit.Blazor/assets/12828888/4f2a4ffc-63a4-475b-b723-fc83fed723cc)
https://app.screencast.com/IYkr97y1W2w1L

PR #867, opened 2024-08-12
#866 [HxAutosuggest, HxSearchBox] Clickable area for reset input is too small, #604 [HxTreeView] Collapse area smaller than expand

petrhank:


Havit.Blazor.Components.Web.Bootstrap/Forms/Autosuggests/Internal/HxAutosuggestInternal.razor
```diff
@@ -31,27 +31,27 @@
...
-                else if (SearchIconEffective is not null)
-                {
-                    <HxIcon CssClass="hx-autosuggest-input-icon-search" Icon="@SearchIconEffective" />
-                }
-            </div>
+            }
+            else if (SearchIconEffective is not null)
+            {
+                <HxIcon CssClass="hx-autosuggest-input-icon hx-autosuggest-input-icon-search" Icon="@SearchIconEffective" />
+            }
```
hakenr, line 53: However, this change removes the wrapping `div.hx-autosuggest-input-icon` that was around **any** icon, not just _Clear_. This alters the structure, so that some component icons have a `div` around them, while others don’t (and have `hx-autosuggest-input-icon` directly on them). This doesn't seem consistent to me.

If we need to add a handler to the wrapping `div`, I would either create a single `div` and one `HxIcon` with everything controlled by conditions, or I would create several branches with `div` + `HxIcon`, ensuring that each branch has a consistent structure of `div` + `HxIcon`.

@crdo?

hakenr (changes requested):
Next time, if there are two issues (#866 and #604), two components, or two separate ways to solve the issue, and the code changes do not overlap, please create two separate commits (and preferably two PRs).

Havit.Blazor.Components.Web.Bootstrap/Forms/Autosuggests/Internal/HxAutosuggestInternal.razor
crdo: @hakenr @petrhank the wrapper is there because of the positioning and should be around all of the icons. I have no preferece between single div or wrapp all, let`s just be consistent and do it same way in SearchBox and Autosuggest.
I also prefere naming the handler ClearInputAsync over HandleCrossClick.

petrhank:
I chose to use the second suggested solution as this is already done this way in HxSearchBox.

hakenr (approved):
lgtm

Issue #872, opened 2024-08-26
[HxInputBase] Allow for CustomValidation Context and Propety and Add LabelType into application-wide Defaults


mmonteagudo:
- Adds the EditContext, ValidationFor and ValidationForFieldName (as implemented in HxValidationMessage) in order to have built-in custom validation in all Input components
- Adds LabelType as an option in the InputBase.Defaults to set application-wide LabelType for Input components that apply.

mmonteagudo:
I already have the code updated, but I can't push the contribution or even create the PR, let me know how can I help

hakenr:
Hello @mmonteagudo,

I like the new `LabelType` settings. I'll just polish the resolution cascade to align with the other parameters.

Could you provide an example scenario where you would take advantage of the new `ValidationXy` parameters?

mmonteagudo:
I want able to provide it for the examples in the code, but in my case I'm using it for Validating using FluentValidation when validating values that are not parameters, for example in my case:

One if my properties in my project is ApplicationData which is a Dictionary, and I need to validate when one for the items in the dictionary has an specific value as a condition to validate all others, as default validation only works with properties, that's not possible, so in this in case I declared in Fluent Validation a "custom property to validate" and I'm matching that property (in this case equals to the dictionary key)

So, basically I'm sending to Fluent a custom property name to validate.

Hope the explanation helps, here are some code examples where I'm using the implementation

![image](https://github.com/user-attachments/assets/0fea7dfc-51e1-414f-8337-27d90ca28494)

![image](https://github.com/user-attachments/assets/4907ac82-346e-4ba0-a9c0-02754cbde400)

hakenr:
Hello @mmonteagudo, I revisited your scenario and from my perspective, the main issue you're trying to resolve is:

* You're binding to dictionary items, which is inherently tricky, and you're using a custom `ValueChanged` callback to capture the value and save it to the dictionary.
* You want to combine this with the built-in `HxValidationMessage` and convince the input component to display the validation message for an artificial property.

From my point of view, your approach works, but it feels like a bit of a hack to solve a very niche scenario. Maintaining such a specific implementation might bring more complexity than the value it offers.

My suggestion for anyone facing a similar situation would be:
1. Try to avoid binding to the dictionary. Instead, create a ViewModel with properties for each individual item.
[...]

PR #873, opened 2024-08-26
[HxInputBase] Allow for CustomValidation Context and Propety and Add LabelType into application-wide Defaults

mmonteagudo:
Linked Issue
#872


mmonteagudo:
https://github.com/havit/Havit.Blazor/issues/872

Havit.Blazor.Components.Web.Bootstrap/Forms/InputSettings.cs
```diff
@@ -9,4 +9,8 @@ public record InputSettings
 	/// Specifies how the validation message should be displayed.
 	/// </summary>
 	public ValidationMessageMode? ValidationMessageMode { get; set; }
+	/// <summary>
+	/// Specifies how the label should be displayed.
+	/// </summary>
+	public LabelType? LabelType { get; set; }
```
hakenr, line 15: The current issue is that not all input components have `LabelType` (implement `IInputWithLabelType`), which is why we left this unimplemented and didn't allow setting `LabelType` from a single place.

However, I understand there's a valid need to have a central point to set `LabelType`, even if some input components don't support it. I'll do some rework here...

Havit.Blazor.Documentation/Pages/Components/FormInputs/FormInputs_Demo_CustomValidation.razor
```diff
@@ -0,0 +1,65 @@
+﻿<EditForm Model="@_testField" OnValidSubmit="@(() => model.Text = "Validation passed... :-)")">
+	<DataAnnotationsValidator />
+
+    <HxInputText Label="HxInputText" @bind-Value="@_testField.TestText" ValidationEditContext="new(_testField.TestModel)" ValidationFor="() => _testField.TestModel.Text"
+        Placeholder="Enter text here." Hint="Hint: Enter text, the value is shared with all text inputs." />
```
hakenr, line 5: This seems a bit unclear:

* The current `<EditForm />` uses `_testField` as the `Model`, while the inputs bind to `@model` properties. Why not use `<EditForm Model="@model">` instead?
* The new `ValidationXy` will cause the form to display a validation message that the user cannot resolve. Is there no way to set `_testField.TestModel.Text`? Additionally, the `HxInputText` will still trigger validation on the bound property, even if the `HxValidationMessage` is redirected elsewhere.
* Have you considered using nested `<EditForm />` components for the different models used by `HxInputText` in this context?
* Have you thought about using `<HxValidationMessage />` directly, combined with `<HxInputText ValidationMessageMode="ValidationMessageMode.None" />` for inputs where you don't want to expose their validation messages?

@mmonteagudo I might not fully understand the intent here. Could you provide a real-world scenario to clarify?

mmonteagudo: I want able to provide it for the examples in the code, but in my case I'm using it for Validating using Fluent Validation when validating values that are not parameters, for example in my case:

One if my properties in my project is ApplicationData which is a Dictionary, and I need to validate when one for the items in the dictionary has an specific value as a condition to validate all others, as default validation only works with properties, that's not possible, so in this in case I declared in Fluent Validation a "custom property to validate" and I'm matching that property (in this case equals to the dictionary key)

So, basically I'm sending to Fluent a custom property name to validate.

Hope the explanation helps, here are some code examples where I'm using the implementation

image

image

hakenr: Custom validation feature removed from the PR.

hakenr:
Input components (`HxInputBase` descendants) that do not support `LabelType`:
* `HxCheckbox`, `HxSwitch`
* `HxCheckboxList`, `HxRadioButtonList`
* `HxInputDateRange`
* `HxInputRange`
* `HxInputSelect` - can we add support here?

...Adding `LabelType` to `InputSettings` will introduce this property to the settings/defaults of all inputs, which could lead to some confusion. I'm considering an alternative approach by adding a brand new ambient defaults such as `HxSetup.Defaults.LabelType` and `HxSetup.Defaults.InputSize`. I'll discuss this with my colleagues.

hakenr (approved):

hakenr:
@mmonteagudo Thanks a lot for your contribution! It made me further develop the idea of "global defaults for all inputs," and I’ve added `HxSetup.Defaults` for these global settings, which currently allow setting `LabelType` and `InputSize`.

Apologies for not accepting the custom validation parameters idea. I’m concerned that this feature could introduce more issues than the value it adds. We might revisit this later, but for now, I’ve decided against it, as there are possible workarounds that can help you achieve the same goal without the proposed changes.

mmonteagudo:
> @mmonteagudo Thanks a lot for your contribution! It made me further develop the idea of "global defaults for all inputs," and I’ve added `HxSetup.Defaults` for these global settings, which currently allow setting `LabelType` and `InputSize`.
> 
> Apologies for not accepting the custom validation parameters idea. I’m concerned that this feature could introduce more issues than the value it adds. We might revisit this later, but for now, I’ve decided against it, as there are possible workarounds that can help you achieve the same goal without the proposed changes.

No worries, I understand.

That's an important feature for our implementation, so we are using it from our fork of Havit, so we just keep it up to date with latest updates and deploy a new package in our private package manager.

I do have a couple of customizations that works for our specific implementation, so I understand. :), Thanks for accepting the other change, that one was pretty important too.

Issue #914, opened 2024-10-30
[Doc] Canonical URLs metadata


hakenr:
Google Search Console is flagging that canonical URLs are missing on our pages.

Following [Google's guidelines](https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls?sjid=6686017794252943981-EU#rel-canonical-link-method), we should add the `rel="canonical"` link metadata to our documentation pages.

The canonical URL format should be:
`https://havit.blazor.eu/components/HxButton` (preserving the specified casing).

PR #915, opened 2024-10-30
[Doc] Canonical URLs metadata #914

Harvey1214:


hakenr:
We need this to cover most pages (if not all). See the list of duplicates identified by Google Search Console.  
Consider using a root cascading value to manage the registrations. I think the first one should always take precedence (no need for `MainComponent`). We can add a special `<PageCanonicalUrl />` component to act as the initial registrant in cases where the order of components on the page doesn't satisfy the rule.

[Duplicate URLs without canonical.txt](https://github.com/user-attachments/files/17575166/Duplicate.URLs.without.canonical.txt)

hakenr (changes requested):
see my previous comment

hakenr (approved):
consolidated, merged master, lgtm

PR #930, opened 2024-11-05
Automatic redirect to premium content

Harvey1214:


Havit.Blazor.Documentation/Havit.Blazor.Documentation.csproj
```diff
@@ -16,6 +16,7 @@
 	<ItemGroup>
 		<PackageReference Include="LoxSmoke.DocXml" />
 		<PackageReference Include="Microsoft.AspNetCore.Components.WebAssembly" />
+		<PackageReference Include="Microsoft.AspNetCore.Http.Abstractions" />
```
hakenr, line 19: > This package has been deprecated as it is legacy and is no longer maintained.

hakenr (approved):
lgtm, I will fix the package issue on my own

Havit.Blazor.Documentation/Havit.Blazor.Documentation.csproj
hakenr: @Harvey1214 See the commit added. The solution is a little tricky as we don't want to add any extra dependency to WASM (downloaded to browser).

hakenr (approved):

Issue #532, opened 2023-05-04
[HxSidebar] Sidebar remains collapsed horizontally when the window is resized 


TPIvan:
When a Sidebar is horizontally collapsed by toggler and then the window is made narrower which causes the Sidebar to collapse vertically to its header and then the Sidebar is expanded vertically by the hamburger expader it remain collapsed horizontally. There is no way how to expand the sidebar horizontally in this situation (the toggler is hidden).
![image](https://user-images.githubusercontent.com/39948593/236154710-8e8d8be0-2660-4421-8da4-4dc63a82888c.png)

### Steps to reproduce
1. Open the documentation site https://havit.blazor.eu/
2. Toggle the Sidebar 
3. Make the window narrower to make Sidebar to collapse to up
4. Expand the Sidebar

### Solution
A solution is probably quite simple the navbar toggler (hx-sidebar-navbar-toggler) should expand the sidebar also horizontally.
The onclick is added:

              <HxButton OnClick=@(async () => await SetCollapsed(false))
                    CssClass="@CssClassHelper.Combine("hx-sidebar-navbar-toggler ms-auto", GetResponsiveCssClass("d-??-none"))"

and the function:
       
       private async Task SetCollapsed(bool state)
	    {
		Collapsed =state;
		await InvokeCollapsedChangedAsync(Collapsed);
	    }

hakenr:
@crdo Can you please check the proposed solution?

crdo:
That is unfortunately not enough. 
We need to get rid of the `collapsed` class on `HxSidebar` and if this is suppose to be a real solution, it has to happen on the window resize event and also take in an account the responsive settings of the Sidebar.
That being said makes it a way more complicated than the proposed solution

TPIvan:
The presence of `collapsed ` css class is controlled by `Collapsed` property.

        private string GetCollapsedCssClass() => Collapsed ? "collapsed" : null;
and onclik handler sets the property whenever the hamburger button is pressed. 
Afte the window resize sidebar is **horizontally** collapsed  (disapears) and shows only after clicking the hamburger button which expands it **horizontally** too and there is no need for reaction to window resize.   

I apologize - it is difficult to explain it clearly.

crdo:
I think there is a better solution. Working on it...

crdo:
@hakenr i decided to link WIP PR as I am not making any progress.

1) cleaned some CSS and moved utility classes to CSS isolation
[...]

PR #943, opened 2024-11-19
[HxSidebar] responsive collapsing

### LINKED ISSUE #282: [HxSidebar] [doc?] Sidebar width in demos
There is `--hx-sidebar-width` CSS variable (with 250px default, but set to 320px for our documentation site).
Despite this setting, the sidebar demos are currently much wider (see the expand arrow on the right).
image

As far as I remember, the width setting used to apply for the demos.
Probably something changed in the HTML/CSS structure that it does not now.
Do we need some sort of container/flex for the sidebar to apply?
Should this be part of sidebar, or how to make this work everywhere?

cc @crdo

crdo:
Fixes #282 and #532.

What i did to fix https://github.com/havit/Havit.Blazor/issues/532 
To hide the item content when the sidebar is collapsed, instead of conditional rendering with if statement, i am adding css utility classes that only hide the content on viewport larger than Sidebars ResponsiveBreakpoint setting and if resized below, it gets displayed again.

@hakenr I need two things from you to be able to complete this PR:
1. Review the API of HxSidebarItemNavLinkContent (naming, etc.)
2. Extract GetResponsiveCssClass somewhere so it is not duplicated (HxSidebar, HxSidebarItem, HxSidebarBrand) and it can be used on other places as I find it useful.

I have also overhauled Sidebars CSS vars so you can create color variations with utility classes and added respective demos and added flexbox layout to implement what was requested in #282.


Havit.Blazor.Documentation/Pages/Components/HxSidebarDoc/HxSidebar_Demo_Colors.razor.css
hakenr, line 1: @crdo Demos can't go beyond the basic `.razor` file, otherwise you'll need to use the `<CodeSnippet />` component to show any additional files.
I'll include the CSS definitions in the `.razor` file with some notes explaining that they're for demonstration purposes.

hakenr (approved):

PR #991, opened 2025-01-21
To hide icon while button is spinning instead of removing from DOM

crdo:
To resolve an issue where tooltips hang on displayed when button is clicked while hovering the icon, we hide the icon instead of remove from DOM.


hakenr (approved):
lgtm

PR #994, opened 2025-01-24
Strong api string localizers generator using incremental generator

jirikanda:
+ No longer generates marker classes (no garbage in IntelliSense 💪).
+ No longer uses DelegatingStringLocalizer (Havit.Extensions.Localization now has no purpose).
+ Generates neutral value comments in the code.
+ No longer searches for or uses the ResourceServiceCollectionInstaller partial class (the generator now produces its own static class).
+ Includes a unit test.

- Localizers no longer implement IStringLocalizer<>, they implement only IStringLocalizer (therefore, no marker class is needed).
- ~~Generators cannot detect RESX files unless explicitly mentioned in the .csproj file.~~
- ~~Generates a warning when no RESX file is found, including a hint on how to modify the .csproj file (can check if this can be implemented using a code fix).~~
- Uses a clone of HttpUtilityExt for human-readable, culture-specific character HTML encoding (e.g., "á" remains "á" and is not converted to an HTML entity).

+ Does not depend on the latest version of Microsoft.CodeAnalysis.* (not 4.12.0, but 4.8.0). The reason is to be able to well used with other nugets. Ie. [Microsoft.EntityFrameworkCore.Design](https://www.nuget.org/packages/Microsoft.EntityFrameworkCore.Design#dependencies-body-tab) depends on [Microsoft.CodeAnalysis.Workspaces.MSBuild](https://www.nuget.org/packages/Microsoft.CodeAnalysis.Workspaces.MSBuild/) and [Microsoft.CodeAnalysis.CSharp.Workspaces](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Workspaces/) 4.8.0+. BUT these libraries have other dependecies on a single version (check ie on [Microsoft.CodeAnalysis.CSharp.Workspaces](https://www.nuget.org/packages/Microsoft.CodeAnalysis.CSharp.Workspaces/#dependencies-body-tab). So if we use latest version, user of our library, who also uses EF Core, must manually specify version of Microsoft.CodeAnalysis.Common, Microsoft.CodeAnalysis.CSharp.Workspaces, Microsoft.CodeAnalysis.CSharp.Workspaces.Common and Microsoft.CodeAnalysis.Workspaces.MSBuild to the latest version.


BlazorAppTest/BlazorAppTest.csproj
hakenr, line 1: Waiting for further investigation whether this is needed.

hakenr (commented):
lgtm, waiting for the `<AdditionalFiles Include="**\*.resx" />` investigation

BlazorAppTest/BlazorAppTest.csproj
jirikanda: It looks MS looks for resx file using the same method.

See this [line](https://github.com/dotnet/roslyn-analyzers/blob/8fe7aeb135c64e095f43292c427453858d937184/src/Microsoft.CodeAnalysis.ResxSourceGenerator/Microsoft.CodeAnalysis.ResxSourceGenerator/AbstractResxGenerator.cs#L35) in [Microsoft.CodeAnalysis.ResxSourceGenerator](https://github.com/dotnet/roslyn-analyzers/tree/8fe7aeb135c64e095f43292c427453858d937184/src/Microsoft.CodeAnalysis.ResxSourceGenerator)

jirikanda: Solved by just referencing Microsoft.CodeAnalysis. 🤷‍♂️

hakenr: For the record: The `<AdditionalFiles .../>` is needed for `<ProjectReference />` (in `Havit.Blazor.sln`). It is no longer needed for consuming apps referencing via `<PackageReference />`.

hakenr (approved):
lgtm

PR #1000, opened 2025-01-30
HxModal onAfterRenderTasksQueue for show and hide

alexax578:
Analogous implementation to HxOffcanvas - fixes issues where ShowAsync and HideAsync are called in quick succession.


hakenr (approved):
lgtm

PR #1038, opened 2025-03-11
HxPager - throw an exception when TotalPages is less than 1

Harvey1214:
Lukáš Michl informed me about a behavior of `HxPager` that doesn't seem optimal. Currently, when `TotalPages` is set to a value lower than 1, the component doesn't look user-friendly and throws an exception when any button is pressed.

image

In this PR, a mechanism was added that throws an exception when `TotalPages` is set to a value lower than 1, so developers immediately notice that the value passed to this parameter must be checked and modified.

`HxGrid` solves this problem by not rendering the pager at all when `TotalPages` is lower than 2.

### DIFF
--- a/BlazorAppTest/Pages/HxPager_NoItems_Test.razor
+++ b/BlazorAppTest/Pages/HxPager_NoItems_Test.razor
@@ -1,0 +1,16 @@
+@page "/HxPager_NoItems_Test"
+
+<h1>HxPager NoItems Test</h1>
+
+<HxButton Color="ThemeColor.Primary" Text="Change TotalPages to 0" OnClick="ChangeTotalPagesToZero" CssClass="mb-2" />
+
+<HxPager CurrentPageIndex="0" TotalPages="_totalPages" />
+
+@code {
+	private int _totalPages = 1;
+
+	private void ChangeTotalPagesToZero()
+	{
+		_totalPages = 0;
+	}
+}
--- a/Havit.Blazor.Components.Web.Bootstrap/Grids/HxPager.razor.cs
+++ b/Havit.Blazor.Components.Web.Bootstrap/Grids/HxPager.razor.cs
@@ -123,10 +123,8 @@ public partial class HxPager : ComponentBase
 	{
 		if (TotalPages < 1)
 		{
-			throw new InvalidOperationException($"{nameof(TotalPages)} has to be greater than or equal to 1.");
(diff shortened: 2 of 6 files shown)


BlazorAppTest/Pages/HxPager_NoItems_Test.razor
hakenr, line 1: We do not use `BlazorAppTest` any more. All the new tests belong to `TestApp.Client`.

Havit.Blazor.Components.Web.Bootstrap/Grids/HxPager.razor.cs
```diff
@@ -118,6 +119,16 @@ static HxPager()
 	[Parameter] public string CssClass { get; set; }
 	protected string CssClassEffective => CssClass ?? GetSettings()?.CssClass ?? GetDefaults().CssClass;
 
+	protected override void OnParametersSet()
+	{
+		if (TotalPages < 1)
+		{
+			throw new InvalidOperationException($"{nameof(TotalPages)} has to be greater than or equal to 1.");
```
hakenr, line 126: Few commits ago we started to prefix "all" exceptions with a component name (except the exceptions regarding missing defaults). See the other `throw`s in the project.

Havit.Blazor.Components.Web.Bootstrap/Grids/HxPager.razor.cs
```diff
@@ -118,6 +119,16 @@ static HxPager()
...
+		}
+
+		base.OnParametersSet();
```
hakenr, line 129: (Blazor documentation suggests not to call empty `base` lifecycle methods when you inherit directly from `ComponentBase`.)

hakenr (changes requested):

Havit.Blazor.Components.Web.Bootstrap/Grids/HxPager.razor.cs
hakenr: This still does not align with the new standard for exceptions
1. The component name should be taken from `this.GetType().Name` to reflect inheritance (if someone derives a new component from our component, we want the exception to report the real component name).
2. The typography is meant to be `[HxPager] Some exception text.`.

```csharp
throw new InvalidOperationException($"[{GetType().Name}] Some exception text...");
```

PR #1043, opened 2025-03-12
Add GrpcWebHandler factory parameter to AddGrpcClientInfrastructure methods

alexax578:
Added an optional parameter to provide custom factory method for the **GrpcWebHandler**.

In my case I needed to provide custom **HttpMessageHandler** to the **GrpcWebHandler**.

https://learn.microsoft.com/en-us/dotnet/maui/data-cloud/local-web-services?view=net-maui-9.0#bypass-the-certificate-security-check


hakenr (approved):
simplified to `TryAddTransient` to allow application to provide its own GrpcWebHandler (factory)

PR #1060, opened 2025-04-07
ECharts

hakenr:


alexax578:
Examples in the documentation look bad in dark mode:
image

alexax578:
Syntax highlighting in the documentation does not seem to work properly:
image

Havit.Blazor.Components.Web.ECharts/HxEChart.razor.cs
```diff
@@ -0,0 +1,151 @@
...
+/// Component for convenient rendering of Apache ECharts.
+/// </summary>
+public partial class HxEChart : IAsyncDisposable
+{
+	private static readonly JsonSerializerOptions s_JsonSerializerOptions;
+
+	/// <summary>
+	/// Unique identifier for the HTML element representing the chart.
+	/// </summary>
+	[Parameter] public string ChartId { get; set; } = $"hx-echart-{Guid.NewGuid()}";
```
alexax578, line 17: Is it worth exposing this Parameter at this time?

The component does not properly handle the case when ChartId changes - when I tested it the auto resize functionality in the basic demo broke...

hakenr: You don't want to change the value, you want to be able to find the instance of ECharts in DOM to be able to do some manual adjustments/interactions.
I find this "safe enough" :-D

hakenr:
> Syntax highlighting in the documentation does not seem to work properly: image

Not part of this PR, known issue of the highlighter we are using.
(Will try an update in separate thread.)

hakenr:
@crdo I will merge this for now to be able to publish the NuGet package. Feel free to add comments, if you find anything. Any code changes will need to go to `master` branch!

Arnnes:
I noticed an issue with clickable axis labels. In the demo `HxEChart_Demo_OnClick`, when `TriggerEvent = true` is set for the `XAxis` and I click on "Mon", the JSON parser fails to parse the value as a decimal because the chart assigns the string "Mon" to params.value.

PR #1075, opened 2025-05-06
[HxGrid] removal of CTS disposal + tests

hakenr:
@jirikanda Please review. If you're good with the change, I'll go ahead and update a few other components that currently dispose the CTS (`HxAutosuggest`, `HxSearchBox`, `HxInputTags`, etc.).

Also, we could try setting `_paginationRefreshDataCancellationTokenSource` to `null` once the data load operation finishes (both QuickGrid and FluentUI-DataGrid do this so the GC can collect the CTS ASAP), though it probably won't make much of a difference.

Reference implementations:
* https://github.com/dotnet/aspnetcore/blob/0508333c3452c8064e8d1092de9de29ad3b63bf5/src/Components/QuickGrid/Microsoft.AspNetCore.Components.QuickGrid/src/QuickGrid.razor.cs#L150
* https://github.com/microsoft/fluentui-blazor/blob/54f2645210d9c6fedf57b3afe6c849be407f24ba/src/Core/Components/DataGrid/FluentDataGrid.razor.cs#L369


jirikanda:
I guess NOT OK, see line 730 in HxGrid.razor.cs

Issue #787, opened 2024-04-11
[HxInputDate] [HxInputDateRange] Changing value causes two calls of EditContext.OnFieldChanged


jirikanda:
## Reprosteps
* Handle EditContext.OnFieldChanged event
* Click on a date in the calendar.
* EditContext.OnFieldChanged event handler is called twice

## Why?
(both HxInputDate&lt;TValue&gt; and also HxInputDate**Internal**&lt;TValue&gt; inherit from InputBase&lt;TValue&gt;)

### First call
* HxInputDate**Internal**&lt;TValue&gt;.HandleValueChanged is called.
* HxInputDate**Internal**&lt;TValue&gt;.HandleValueChanged sets the CurrentValueAsString.
* CurrentValueAsString setter sets InputBase&lt;TValuegt;.CurrentValue.
* InputBase&lt;TValue&gt;.CurrentValue calls ValueChanged (important to understand the the second call)
* The InputBase&lt;TValue&gt;.CurrentValue setter calls EditContext.NotifyFieldChanged for the first time.

### Second call
* HxInputDate&lt;Value&gt; handles ValueChanged event
* It sets the CurrentValue property (see BuildRenderInputCore method, sequence number 101).
* The InputBase&lt;TValue&gt;.CurrentValue setter calls (again) EditContext.NotifyFieldChanged.

### Posible solutions (not sure right now)
* only  HxInputDate&lt;TValue&gt; should derive from InputBase&lt;TValue&gt;
* OR HxInputDate&lt;TValue&gt; should not Set CurrentValue, but it should fire ValueChanged event

(I guess this occurs also in HxInputDateRange while it uses the same code structure.)

hakenr:
> only HxInputDate<TValue> should derive from InputBase<TValue>

My preference (same in other similar components).

jirikanda:
@hakenr, I am not familiar with the current code structure, there for I am not able to fix the issue (until someone explains me the thoughts behind). 
ie.
```
	private void HandleValueChanged(string newInputValue)
	{
#if NET8_0_OR_GREATER
		CurrentValueAsString = newInputValue;
#else
		// HandleValueChanged is used instead of TryParseValueFromString
		// When TryParseValueFromString is used (pre net8), invalid input is replaced by previous value.		
		bool parsingFailed;
		_validationMessageStore.Clear(FieldIdentifier);

[...]

PR #1076, opened 2025-05-12
[HxInputDate] [HxInputDateRange] Changing value causes two calls of EditContext.OnFieldChanged #787

jirikanda:
HxInputDate
- HxInputDateInternal now forwards value handling to HxInputDateInternal via CurrentValueAsStringChanged.
- an enclosing `if (FieldIdentifier.Model != null)` in the HxInputDateInternal.razor.cs was removed

HxInputDateRange
- handling from/to date parts stays in HxInputDateRangeInternal
- only (the whole range) value changes are propagated from HxInputDateRangeInternal to HxInputDateRange
- HxInputDateRangeInternal still calls `EditContext.NotifyFieldChanged(_fromFieldIdentifier/_toFieldIdentifier)` (functionality not changed), this means any valid change fires NotifyFieldChanged twice (_fromFieldIdentifier/_toFieldIdentifier and the "whole" value)
- order of actions in methods calling CurrentValueChanges is perserved but can be subject of futher discussion (ie. should `HandleToClearClickAsync` method call first `CurrentValueChanged` or `CloseDropdownAsync`?)

HxInputDateInternal and HxInputDateRangeInternal does not implement IDisposable anymore (but still implements IAsyncDisposable).

### DIFF
--- a/Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateInternal.razor.cs
+++ b/Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateInternal.razor.cs
@@ -69,7 +69,7 @@ public partial class HxInputDateInternal<TValue> : ComponentBase, IAsyncDisposab
 
 	[Parameter] public string NameAttributeValue { get; set; }
 
-	[Parameter(CaptureUnmatchedValues = true)] public IReadOnlyDictionary<string, object> AdditionalAttributes { get; set; }
+	[Parameter] public IReadOnlyDictionary<string, object> AdditionalAttributes { get; set; }
 
 	[Inject] protected IStringLocalizerFactory StringLocalizerFactory { get; set; }
 	[Inject] protected IJSRuntime JSRuntime { get; set; }
--- a/Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateInternal.razor.cs
+++ b/Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateInternal.razor.cs
@@ -90,7 +90,6 @@ public partial class HxInputDateInternal<TValue> : ComponentBase, IAsyncDisposab
 
 	private async Task HandleCurrentValueAsStringChanged(string newInputValue)
 	{
-		//CurrentValueAsString = newInputValue;
 		await CurrentValueAsStringChanged.InvokeAsync(newInputValue);
 	}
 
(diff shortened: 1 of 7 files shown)


Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateInternal.razor.cs
```diff
@@ -63,6 +67,10 @@ public partial class HxInputDateInternal<TValue> : InputBase<TValue>, IAsyncDisp
 
 	[Parameter] public DateTime CalendarDisplayMonth { get; set; }
 
+	[Parameter] public string NameAttributeValue { get; set; }
+
+	[Parameter(CaptureUnmatchedValues = true)] public IReadOnlyDictionary<string, object> AdditionalAttributes { get; set; }
```
hakenr, line 72: I think the internal component should not use `CaptureUnmatchedValues = true` as the `HxInputDate` component should set the `AdditionalAttributes` parameter explicitly?

hakenr: I.e.
```diff csharp
-builder.AddMultipleAttributes(300, AdditionalAttributes);
+builder.AddAttribute(300, nameof(HxInputDateInternal<TValue>.AdditionalAttributes)), AdditionalAttributes);
```

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateRangeInternal.razor.cs
```diff
@@ -43,6 +51,8 @@ public partial class HxInputDateRangeInternal : InputBase<DateTimeRange>, IAsync
 
 	[Parameter] public IconBase CalendarIconEffective { get; set; }
 
+	[Parameter(CaptureUnmatchedValues = true)] public IReadOnlyDictionary<string, object> AdditionalAttributes { get; set; }
```
hakenr, line 54: dtto

Havit.Blazor.Components.Web.Bootstrap/Forms/Internal/HxInputDateInternal.razor.cs
hakenr: (I understand the original `AdditionalAttributes` came from `InputBase` with `CaptureUnmatchedValues`, but since this is an internal component, we can take advantage of the fact that it's only called from `HxInputDate`, where the `AdditionalParameters` parameter can be passed in its entirety.)

hakenr (approved):

Issue #794, opened 2024-04-22
[HxInputDate] - When selected date year is out of the MinDate-MaxDate, year is not displayed


jirikanda:
* Current date is in 2023.
* Min&MaxDates are in 2024.

https://github.com/havit/Havit.Blazor/assets/5111719/4606465b-b556-4f60-babf-bcfb30184e17

hakenr:
@jirikanda @crdo What do we expect the component to do?
* throw an exception on invalid parameters?
* override the Min/Max parameters to include current value?

hakenr:
> > throw an exception on invalid parameters?
> 
> From my perspective, everything is valid.
> 
> * User (in the past) entered some value which is saved in the database.
> * We are showing user's value and let the user fix the value to the current min-max date range.

Not entirely sure, but you are probably right.

The current documentation/specification states _"The first/last date selectable from the dropdown calendar."_ for the `MinValue`/`MaxValue` parameters. We could allow the `Value` to be outside these bounds, but the dropdown calendar will still only allow selections within the `MinDate..MaxDate` range (as far as I know, we don't use these boundaries for validation; it's up to the user to validate the value in the model). So technically, we can open a calendar dropdown for the month of the currently selected value, but still with "disabled" values (strikethrough).

Issue #793, opened 2024-04-22
[HxInputDate] - Navigation in calendar with MinDate&MaxDate - previous month click skipped, next month returns to the first month


jirikanda:
* HxInputDate value is limited by MinDate and MaxDate
* Click on the next month
* Click on the previous month
* The selected month does not change (bug)
* Click on the next month until the last month is selected.
* Click on the next month, first month is selected (bug)

https://github.com/havit/Havit.Blazor/assets/5111719/a8e5915b-62dc-42a3-bac8-52c9b364def2

hakenr:
"Value of the HxInputDate is out of the MinDate-MaxDate range".
Lets resolve the #794 first (we might come to conclusion, that we should throw exception on invalid combination of parameters and this issue becomes obsolete then?)

MBNSoftware:
To me, if #794 is resolved, then this one will become obsolete. Unless there's something else about this bug, of course :)

Since validation is on the end-user side, technically you would only need to change the month that is displayed according to previous/next buttons. Invalid months would be striked-through, indicating an out-of-bounds date.

PR #1078, opened 2025-05-13
[HxInputDate] - When selected date year is out of the MinDate-MaxDate, year is not displayed

jirikanda:
* DisplayMonth is re-evaluated everytime Value is changed by the parameter ("from outside").
* When Value is set (not null), DisplayMonth is not limited by Min/MaxDateEffective anymore.
* When Value is null, DisplayMonth is set to current month limited by Min/MaxDateEffective (not changed)
* Year selection contains DisplayMonth.Year even it is out of the Min/MaxDateEffective range.
* Move to previous/next also works in the direction to the Min/MaxDateEffective.

https://github.com/user-attachments/assets/a5a70830-67d6-4452-a0fe-ae19ad24f63b


jirikanda:
Fixes #794 and (if I understand myself after a year it fixes also) #793

Issue #857, opened 2024-07-15
Consider removing FocusAsync method from HxInputBase


jirikanda:
* Now, HxInputBase.FocusAsync focuses InputElement when it is set or throws exception if it is not set.
* InputElement is set only in the descenants (in BuildRenderTree by AddElementReferenceCapture).
* InputElement has no other reason.
* InputElement is a protected property.

Issues:
* The descendants do not know they should (must?) set value to the protected InputElement property.
* The exception message is partialy wrong (they notice only part "focusing too early", not noticing "or InputElement is not set")

Suggestion:
* Remove FocusAsync and InputElement property from HxInputBase and move FocusAsync method only to the components implementing it.
* Or force implementations (HxInputBase descendant) to set InputElement.

FYI, @Arnnes

PR #1081, opened 2025-05-20
#857 Consider removing FocusAsync method from HxInputBase

jirikanda:
Use it or toss it.

Contains breaking changes: 
* There was removed `InputElement` property from the `HxInputBase`.
* There was removed `FocusAsync` method from the `HxInputBase`.


hakenr (approved):
lgtm

Issue #415, opened 2022-10-26
[HxGrid] Ability to reset grid state (position?) when refreshing data (e.g. after new filter applied)


hakenr:
We should give the developers an easy way to navigate the beginning of the grid after new filtering criteria are applied (no matter whether it is in paging, inifinite-scroll or any other future mode). Maybe something like (new optional parameter)
```csharp
await gridComponent.RefreshDataAsync(resetUserState: true);
```
...which will be used when the Apply button in filter is clicked?

Do we also want to reset sorting or just the position?

(extracted from #412)

hakenr:
From discussion:
* we want to jump to start (is it achievable with `Virtualize`? new instance?),
* we want to keep the sorting,
* we want to keep other potential grid settings (column width, column visibility, ...)

hakenr:
@Harvey1214 This is still marked as `#TBD`, i.e., not specified enough to start the implementation. Let me know if you need this functionality for any of your scenarios, and we can move forward.

ericwscott:
@hakenr is it currently possible to reset paging position when the dataset changes?  This task is to provide an easy way, but are there any workarounds?

hakenr:
@ericwscott The current workaround is to use ` @bind-CurrentUserState="_userState"` and reset the `PageIndex` in `_userState` field.

gdube-nrc:
Just want to point out that this workaround doesn't work for a virtualized grid, can you confirm?

hakenr:
@gdube-nrc Correct, restoring the grid position in a virtualized grid isn't supported. The virtualization depends on the underlying `Virtualize` Blazor component, which doesn't offer this feature.

gdube-nrc:
Okay, thank you. Then I guess the only way would be to use JS Interop to reset the scroll position manually.

PR #1088, opened 2025-05-28
[HxGrid] Ability to reset grid state (position?) when refreshing data (e.g. after new filter applied) - sorting

jirikanda:
Adds sorting reset to #415.


hakenr (approved):
lgtm

Issue #1072, opened 2025-04-30
HERON Showcase - image size does not match


hakenr:
![Image](https://github.com/user-attachments/assets/e219c8e1-fe37-440b-b141-d1c10f6677ed)

PR #1099, opened 2025-06-12
to fix TulipHeron showcase img height

Viliammm:
https://github.com/havit/Havit.Blazor/issues/1072


crdo:
@Viliammm I guess there are some changes that did not make it to the commit. You only removed the old TulipHeron image.

PR #1104, opened 2025-06-18
HxPopover - click trigger doesn't work when set with hover trigger

Harvey1214:


crdo:
@Harvey1214 I guess this can be abandoned...

Harvey1214:
@crdo agreed.

Issue #1103, opened 2025-06-16
[HxInputBase] Add aria-describedby attribute to input targeting its validation message


alexax578:
[aria-describedby attribute documentation](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-describedby)

### Motivation:
This attribute would help with accessibility primarily for screen reader users as it would allow them to easily read error messages associated with the given inputs.

### Description:
Based on a short discussion with @jirikanda we could implement this attribute on `HxInputBase`. Unique ids need to be generated for error messages which would be listed in the `aria-describedby` attribute on the `input` element and supplied as ids to the `HxValidationMessage` elements that contain the text message.

There is a case where I do not want to show validation messages uder the given inputs but instead use the `ValidationSummary` component or list the validations myself. I don't know how I could pair these messages to the given inputs easily. I suggest a new `ValidationMessageMode.AriaDescribedByOnly` (name subject to change) which would render the validation message but make it invisible and not take up any space. The input would still be paired to this validation message which would allow me to support the screen readers in a simple manner while rendering the validation messages elsewhere for the user to see.

PR #1110, opened 2025-07-22
[HxInputBase] Add aria-describedby attribute to input targeting its validation message

jirikanda:
Implements #1103.

Content:
* new ValidationMessageMode.VisuallyHidden
* HxValidationMessage renders id (if provided)
* HxInputBase (HxFormValueComponentRenderer) renders hint element id
* HxInputBase adds aria-describedby attribute to the input, value contains hint element id and validation message element id (aria-describedby can contain multiple element ids separated by space)


Havit.Blazor.Components.Web.Bootstrap/Forms/ValidationMessageMode.cs
```diff
@@ -23,6 +23,11 @@ public enum ValidationMessageMode
 	/// </remarks>
 	Floating = 2,
 
+	/// <summary>
+	/// Validation messages are rendered but visually hidden (e.g., for screen readers).
+	/// </summary>
+	VisuallyHidden = 3,
```
hakenr, line 29: I would go only for `Hidden`, it is similar to `<input type="hidden" .. />` - also "rendered" in DOM, but not visible.

Havit.Blazor.Components.Web.Bootstrap/Forms/ValidationMessageModeExtensions.cs
```diff
@@ -12,6 +12,7 @@ internal static string AsCssClass(this ValidationMessageMode mode)
 			ValidationMessageMode.Regular => "invalid-feedback",
 			ValidationMessageMode.Tooltip => "invalid-tooltip text-truncate",
 			ValidationMessageMode.Floating => "invalid-feedback feedback-floating text-truncate",
+			ValidationMessageMode.VisuallyHidden => "visually-hidden",
```
hakenr, line 15: Should this be someting like `hx-validation-message-hidden`? The above are Bootstrap classes, but this is our own.

jirikanda: See https://getbootstrap.com/docs/5.0/helpers/visually-hidden/

Havit.Blazor.Components.Web.Bootstrap/Forms/ValidationMessageMode.cs
hakenr: @jirikanda Do you also prefer `VisuallyHidden` here?

jirikanda: Yes, but it is not a must-have.

Issue #1005, opened 2025-02-05
[HxInputNumber] Toggle value's sign on [MINUS] key


hakenr:
Whenever I press <kbd>-</kbd> on the keyboard, I want the value's sign to toggle.

Issue #545, opened 2023-05-28
[HxInputNumber] No binding update options


robertmclaws:
`HxInputText` has a `BindEvent` property that can be used to indicate how often the binding should be updated during typing. The `HxInputNumber` control has no such binding, and needs one.

robertmclaws:
Any update on this?

hakenr:
This is not currently a priority for us, but it's a #good-first-issue for anyone interested in contributing. The implementation will be similar to what we have in `HxInputText`.

hakenr:
Duplicates #1115, resolved.

PR #1112, opened 2025-07-22
[HxInputNumber] Toggle value's sign on [MINUS] key

jirikanda:
Implements #1005.

The current implementation contains a difference in event behavior between pasting a value and pressing the minus key.
Pressing the minus key behaves "standardly": the key press triggers the `input` event, and the `change` event is only triggered when the input field loses focus.
In contrast, pasting a value from the clipboard triggers the `change` event immediately.


jirikanda:
Current implementation, while correct, blocks implementation of #545. This uses javascript oninput, while oninput for EventCallback is also required.

hakenr (changes requested):
Unfortunately, #545 made it into production and that scenario is much more important. My original idea was to use `onkeydown`/`onkeyup` event. Any chance to avoid `oninput`?

jirikanda:
Fixed all known issues.

Also fixes #1127 (while it needs onblur from #1005 implemention, it is not in a separate PR).

hakenr (approved):
Maybe we should dispatch the whole event-set (beforeXy etc.) like SmartPaste does (in both cases), but the current implementation seams to work. We will see in production...

PR #1126, opened 2025-08-18
Add Anti-Forgery Token Support to HxInputFile Components

aliastopan:
This PR introduces antiforgery token support to `HxInputFile` and `HxInputFileDropZone`, ensuring secure file uploads in server-side Blazor apps. The feature is optional and disabled by default to avoid breaking changes. A new `StandaloneServer` test project was added to validate the functionality in a realistic server-only Blazor environment.

## Background
The `HxInputFile` component requires `UploadUrl` when uploading file(s). In my use case, I needed to use minimal APIs instead of controllers. When a project uses `app.UseAntiforgery()`, methods like `app.MapPost(/**/)` cannot accept payload without passing the anti-forgery token.

This causes the following exception:
```
Microsoft.AspNetCore.Http.BadHttpRequestException: Invalid anti-forgery token found when reading parameter "IFormFile file" from the request body as form. ---> Microsoft.AspNetCore.Antiforgery.AntiforgeryValidationException: The required antiforgery request token was not provided in either form field "__RequestVerificationToken" or header value "RequestVerificationToken". at Microsoft.AspNetCore.Antiforgery.DefaultAntiforgery.ValidateRequestAsync(HttpContext httpContext) at Microsoft.AspNetCore.Antiforgery.Internal.AntiforgeryMiddleware.InvokeAwaited(HttpContext context) --- End of inner exception stack trace --- 
```

Currently, the simplest workaround is to add `.DisableAntiforgery()` to the minimal API endpoint mapping. However, this workaround is not ideal for production scenarios as it removes CSRF protection.

## Solution
First, I introduced the `IncludeAntiforgeryToken` parameter in `HxInputFileCore.cs`, `HxInputFile.cs`, and `HxInputFileDropZone.razor.cs`:

``` csharp
[Parameter] public bool IncludeAntiforgeryToken { get; set; }
```

Then, I modified the `StartUploadAsync(/**/)` method to accept an optional anti-forgery token:

``` csharp
public async Task StartUploadAsync(string accessToken = null, string antiforgeryToken = null)
{
	Contract.Requires<ArgumentException>(!String.IsNullOrWhiteSpace(UploadUrl), nameof(UploadUrl) + " has to be set.");
	await EnsureJsModuleAsync();
	_filesUploaded = new ConcurrentBag<FileUploadedEventArgs>();
	if (_disposed)
	{
		return;
	}
	await _jsModule.InvokeVoidAsync("upload",
		Id,
		_dotnetObjectReference,
		UploadUrl,
		accessToken,
		MaxFileSizeEffective == long.MaxValue ? null : MaxFileSizeEffective,
		MaxParallelUploadsEffective,
		UploadHttpMethodEffective,
		IncludeAntiforgeryToken ? antiforgeryToken : null);
}
```

The `InvokeVoidAsync(/**/)` now uses the `IncludeAntiforgeryToken` property to determine whether an anti-forgery token should be included. This approach ensures backward compatibility as it's opt-in only.

The `HxInputFileCore.js` handles the anti-forgery token by adding the required header:

``` js
if (antiforgeryToken) {
    request.setRequestHeader('RequestVerificationToken', antiforgeryToken);
}
```

### Benefits
- **Secure by design**: Enables proper CSRF protection for file uploads
- **Backward compatible**: Existing implementations continue to work unchanged
- **Opt-in**: Anti-forgery token support is only enabled when explicitly requested
- **Minimal API friendly**: Works seamlessly with ASP.NET Core minimal APIs that require anti-forgery validation

### Breaking Changes
None. This is a purely additive change that maintains full backward compatibility.

## Testing
Since the `BlazorAppTest` is marked as obsolete and uses a pre-.NET 6 Blazor framework, testing the feature with it is not suitable. `Havit.Blazor.TestApp[.Client]` proved to be tricky to work with as well, since Blazor WebAssembly does not include the `Microsoft.AspNetCore.Antiforgery` namespace. Therefore, I added `Havit.Blazor.TestApp.StandaloneServer`, a pure server-side Blazor app for testing purposes.

- [x] Tested with minimal API endpoints that require anti-forgery tokens
- [x] Tested both `HxInputFile[Core]` and `HxInputFileDropZone` components
- [x] Confirmed that disabling `IncludeAntiforgeryToken` works as before


alexandrhajek:
I'd like to chime in - name of the header that holds the Antiforgery token should be configurable through a parameter.

aliastopan:
@hakenr, thank you for the reply and consideration.

To answer your questions:

1. No, we don’t necessarily need the `IncludeAntiforgeryToken` property. Since the `antiforgeryToken` argument is `null` by default and ideally used as a named parameter to differentiate between `accessToken` and `antiforgeryToken`, `IncludeAntiforgeryToken` can be dropped to remove redundancy.

2. The `StandaloneServer` test project can be removed. I created it as a convenient way of testing since I’m not too familiar with the Blazor Server + WASM client setup. Sorry about that. I did try to adjust my testing to use the `TestApp` + `TestApp.Client` setup, but I ended up declaring `HxInputFile_AntiforgeryTest.razor` in `TestApp` instead of `TestApp.Client` because I needed the `Microsoft.AspNetCore.Antiforgery` namespace (WASM don't have that). That in itself caused a problem where the page wasn’t automatically discovered in `index.razor` of `TestApp.Client`.

I’d like to handle the changes myself, but I don’t think I can fully align with the testing standards/conventions.

Also, @alexax578  mentioned that the header name for the antiforgery token should be configurable through a parameter.

I forgot that we declare:
```
builder.Services.AddAntiforgery(options =>
{
    options.HeaderName = "X-Custom-CSRF-Token";
});
```

aliastopan:
I did some refactoring:

1. I removed the `IncludeAntiforgeryToken` property to avoid redundancy. The `string antiforgeryToken = null` argument now determines whether the `StartUploadAsync` method call requires antiforgery validation.

2. I dropped the `StandaloneServer` project test and used the `Havit.Blazor.TestApp` instead. Though I’m not sure if the test fully aligns with this library’s conventions. You might want to check the `index.razor` in `Havit.Blazor.TestApp.Client` later.

3. I added an `antiforgeryHeader` argument to the `StartUploadAsync` method:

```csharp
public async Task StartUploadAsync(
    string accessToken = null,
    string antiforgeryToken = null,
    string antiforgeryHeader = "RequestVerificationToken")
{
    ...
}
```

I set `"RequestVerificationToken"` as the default argument. Antiforgery validation only runs if `antiforgeryToken` is provided; otherwise, `antiforgeryHeader` is ignored in `HxInputFileCore.js`.

Should I push my changes?

hakenr:
@aliastopan Sounds great. Please only rename the method argument from `antiforgeryHeader` to `antiforgeryHeaderName` and push your changes. Thank you!

PR #1193, opened 2025-12-20
[HxIcon] aria-label not allowed, deprecated

efinder2:
The usage of `aria-label` on `<i>`-element is not allowed. This property needs to be set on the parent element.
To nudge the developers in that direction, the property AriaLabel was deprecated.

Additionally some accessibility problems in the Demo site were fixed.


efinder2:
might need to be postponed to the next major version. Changing a property to obsolete might break project with warning as errors. Merge the changes if suitable

hakenr:
@efinder2 HxIcon is a fairly abstract component, and it doesn't assume that the internal implementation will render it using an `<i>` element (that's specific to the BootstrapIcon implementation). So I wouldn't remove it from HxIcon for now.

PR #1230, opened 2026-03-13
Reports request status error when exception is thrown in grpc request

jirikanda:


Havit.Blazor.Grpc.Server/ServerExceptions/ServerExceptionsGrpcServerInterceptor.cs
```diff
@@ -29,6 +30,16 @@ protected override bool OnException(Exception exception, out Status status)
...
 		}
 
+		// Report error in OpenTelemetry
+		var activity = Activity.Current;
+		if (activity?.Status == ActivityStatusCode.Unset)
+		{
+			activity.SetStatus(ActivityStatusCode.Error);
+#if NET9_0_OR_GREATER
+			activity.AddException(exception); // useful when exception is not reported via logging
+#endif
```
hakenr, line 40: @jirikanda Do we want to AddException only when the ActivityStatusCode was Unset (= we are the ones setting it to Error)?

hakenr (approved):

Issue #1192, opened 2025-12-19
Automated Accessiblity Tests


efinder2:
I suggest to add automated accessiblity tests to the test suite. So new created problems would be caught. The application already uses Playwright. There is a nuget package [axe](https://github.com/dequelabs/axe-core-nuget) wich allows to run this kind of test quite easy.

This library is a wrapper around axe. It's used for example in the edge dev tools:
<img width="1360" height="765" alt="Image" src="https://github.com/user-attachments/assets/9093d569-bef3-43ff-be5d-051b3be42b3d" />

I would implemnt it (may take some time). For start i suggest to put all current issue to the ignore list. So no new accessibility problems are introduced. Old ones may be fixed if there is time at hand.

hakenr:
This sounds like a great contribution idea. Go ahead - we’d really appreciate it.

PR #1396, opened 2026-05-04
Added automated accessiblity tests

efinder2:
I've added the execution of automated accessibility tests via axe as discussed in #1192. This is at the moment only proof of concept. I'm not sure where to run them correctly. Currently they are run in the clean up method. It was an easy way to execute it after every test, so i don't have to copy every test. This may not be the final way. The current implementation causes problems with the termination of the chrome headless shell. They are orphans after the execution. It freezes my Laptop with 16GB of RAM. With 32 on my Desktop the tests run through and are killed after the execution by the IDE. 

I added a list of the found problems.
[axe-results-20260504-194743.md](https://github.com/user-attachments/files/27374649/axe-results-20260504-194743.md)

There are multiple ways to implement them further. I don't recommend to add them directly to the pipeline. There are too many findings to fix them at once. So there needs to be a switch to enable them. In the current implementation this is solved by a compile directive. 
The easiest way would be to copy the tests, add a RunAxe at the end and filter by a TestCategory. An alternative would be to call RunAxe after each test (and in relevant steps in between). This method filters by environment variables if the accessibility tests are activated.
Alternatively we run them  always. For now we could exclude all rules with warnings or suppress the rule for some css selectors if they aren't many of this category.


hakenr:
@copilot resolve the merge conflicts in this pull request

hakenr:
@efinder2 Thank you for your contribution. Please take a look at the Copilot code review findings. Just let me know whether you are willing to polish the PR on your own or if we should take over.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -10,6 +13,34 @@ public abstract class TestAppTestBase : PageTest
...
+	/// </summary>
+	[TestCleanup]
+	public async Task Cleanup()
+	{
+#if (ACCESSIBLITYTESTS)
+		await RunAxe();
+#endif
+
+		await Context.CloseAsync();
+	}
```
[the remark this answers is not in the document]
efinder2: fixed

efinder2:
I solved the issues brought by copilot and found, that i don't need to add the configuration to every csproj. How do you prefer the usage of this tool? 

I would exclude the existing problems from the execution. The test will then pass but new accessiblity problems would cause them to fail. I'd suggest to create issues in github for the ignored violations. a lot of them may be grouped in a shared issue because they have the same root problem.

CONTRIBUTING.md
```diff
@@ -27,4 +28,50 @@ protected virtual ComponentNameSettings GetDefaults() => Defaults;
...
+Build the test project and install the Playwright browsers before running tests.
+```pwsh
+# Build the test project first, then install the browsers.
+# Replace net10.0 with the target framework you are using if different.
+dotnet build Havit.Blazor.E2ETests\Havit.Blazor.E2ETests.csproj
+Havit.Blazor.E2ETests\bin\Debug\net10.0\playwright.ps1 install
+```
+
+### Run accessibility tests
+The Axe execution is enabled by the environment variable `ACCESSIBILITYTESTS`.
```
hakenr, line 46: This is not what the code does. `AccessibilityTestsEnabled` in `TestAppTestBase` is a compile-time `#if ACCESSIBILITYTESTS`, defined only when the csproj is built with `-c ACCESSIBILITYTESTS`. Nothing reads an environment variable.

I verified it: `$env:ACCESSIBILITYTESTS="true"; dotnet test Havit.Blazor.E2ETests\Havit.Blazor.E2ETests.csproj` builds Debug, runs zero axe checks, passes green, and still writes an `axe-results-*.md` saying "No Errors Found". Only `dotnet test -c ACCESSIBILITYTESTS` actually runs axe, and that command appears nowhere in the docs (line 76 even says the opposite). The `bin\Debug\net10.0\playwright.ps1` path on line 42 is also wrong for the configuration that enables axe.

Suggestion: make the code match these docs, i.e. `private static bool AccessibilityTestsEnabled => Environment.GetEnvironmentVariable("ACCESSIBILITYTESTS") is "true" or "1";` and delete the build-configuration plumbing (see my comment on the slnx).

Havit.Blazor.slnx
```diff
@@ -1,5 +1,8 @@
 <Solution>
   <Configurations>
+    <BuildType Name="ACCESSIBILITYTESTS" />
```
hakenr, line 3: A solution-wide build type to flip one `bool` in one test project is too heavy. Because there is no per-project mapping here, `Configuration=ACCESSIBILITYTESTS` is passed to all 38 projects, and even `dotnet test <csproj> -c ACCESSIBILITYTESTS` flows through the `ProjectReference` chain: I ended up with `bin/ACCESSIBILITYTESTS/` output in `Havit.Blazor.TestApp` and `Havit.Blazor.Components.Web.Bootstrap`, built in a configuration none of them declare (no `DEBUG` constant, no `Optimize`). The audit then runs against a third binary flavour that is neither Debug nor Release, VS gets a permanent extra entry in the configuration dropdown, and CI could never reuse its `--no-build -c Release` binaries for an a11y run.

Please drop the slnx change, the `<Configurations>`/`<Platforms>`/`DefineConstants` in the csproj and the `#if`, and read an environment variable at runtime instead (the repo already toggles Playwright via env vars in `PlaywrightInteractiveDebug.runsettings`). `<Platforms>AnyCPU</Platforms>` is a no-op default either way.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -29,4 +71,74 @@ protected async Task NavigateToTestAppAsync(string relativePath)
...
+	protected async Task RunAxe()
+	{
+		AxeRunOptions options = new AxeRunOptions(
+			runOnly: new AxeRunOnly(AxeRunOnlyType.Tag, new List<string>
+			{
+				"wcag2a",
+				"wcag2aa",
+				"wcag21a",
+				"wcag21aa",
+				"EN-301-549"
```
hakenr, line 84: This tag does nothing with the library as referenced. `Playwright.Axe` 1.3.0 (last release July 2022) embeds axe-core **4.4.3**; the `EN-301-549` tags arrived in axe-core 4.8. I checked the package in the NuGet cache: the dll carries the banner `axe v4.4.3` and the string `EN-301-549` appears nowhere in it, so the tag silently matches zero rules and the EN 301 549 coverage advertised in CONTRIBUTING.md does not exist. The whole rule set is 2022 vintage (no WCAG 2.2 rules either).

Please switch to Deque's official `Deque.AxeCore.Playwright` (currently 4.13.0, axe-core 4.13, same `Microsoft.Playwright >= 1.20.2` dependency). The API is nearly the same (`page.RunAxe(AxeRunOptions)`, `AxeRunOnly`, `AxeRuleObjectValue`), and it also ships an HTML report option.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -10,6 +13,45 @@ public abstract class TestAppTestBase : PageTest
...
+	/// - Zero A11y execution unless explicitly enabled
+	/// </summary>
+	private static bool AccessibilityTestsEnabled =>
+#if ACCESSIBILITYTESTS
+		true;
+#else
+		false;
+#endif
+
+	public override async ValueTask DisposeAsync()
```
hakenr, line 35: Design question, which I know you raised yourself: hooking every test's `DisposeAsync` makes the audit unit "whatever DOM state the test happened to leave" rather than "a TestApp page". Concretely, on this branch the 118 `[Fact]`s navigate to only 37 of the 86 `@page` routes in TestApp. `/HxInputDateRange_EditForm` and `/HxInputDateRange_NoEditForm` get scanned 9x each (so one contrast bug produces 9 failures and 9 near-identical report rows), while all `HxGrid_*`, `HxTreeView`, `HxCheckbox*`, `HxEChart_*` pages are never audited. And the only opt-out is the global `rules` dictionary, although you note there are too many findings to fix at once.

I would prefer a dedicated `AccessibilityTests` class with a `[Theory]` + `[MemberData]` over TestApp routes (navigate, wait for `#blazor-ready-for-tests`, `RunAxe`). That covers every page exactly once, lets us `Skip` individual pages while burning findings down, and leaves `TestAppTestBase`, the static `StringBuilder`, the lock and the assembly fixture untouched. Interaction states (open modal, open dropdown) can be added later as explicit a11y tests where they matter.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -10,6 +13,45 @@ public abstract class TestAppTestBase : PageTest
...
+	{
+		try
+		{
+			if (AccessibilityTestsEnabled)
```
hakenr, line 39: If the DisposeAsync approach stays: this runs axe even when the test body already failed. A `WaitForSelectorAsync("#blazor-ready-for-tests")` timeout leaves a half-rendered (or `about:blank` / chrome-error) page; axe then scans it, appends rows for a state the test never intended, and throws a second `Assert.Fail`, so xUnit reports a two-exception aggregate led by the Markdown table and the original Playwright timeout is buried. On `about:blank`, `html-has-lang` (Serious) would even produce a spurious failure.

Playwright's own base classes gate teardown on `TestOk()`; mirror that here:

```csharp
if (AccessibilityTestsEnabled && TestContext.Current.TestState?.Result == TestResult.Passed)
{
	await RunAxe();
}
```

Also note that an exception from `base.DisposeAsync()` in the `finally` (e.g. a crashed browser during `Context.CloseAsync`) replaces the in-flight `Assert.Fail`, so the violation table never reaches the test output. Catch and aggregate if you keep this structure.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -10,6 +13,45 @@ public abstract class TestAppTestBase : PageTest
...
+		}
+		finally
+		{
+			// always let Playwright tear down the browser context (prevents orphaned headless browser processes)
+			await base.DisposeAsync();
+		}
+	}
+
+	// Temporary storage for the axe Result
+	internal static readonly StringBuilder AxeReport = new();
```
hakenr, line 52: This plumbing (static `StringBuilder` + static `Lock` + the fixture's `Dispose` rendering the table header while the rows are formatted here) can go away entirely:

- The lock protects nothing: `AssemblyInfo.cs` sets `CollectionPerAssembly`, so tests run serially, and the assembly fixture disposes after all of them. It mainly signals thread-safety that would be false if someone re-enabled parallelisation (`TestContext.Current` and the shared browser are not safe).
- Header and row format live in two files and will desync the first time someone adds a column.
- `Console.WriteLine` from a fixture `Dispose` is not part of the TRX that CI publishes.

xUnit v3 4.0 has `TestContext.Current.AddAttachment("axe-violations.md", text)`, which flows into the TRX we already produce (`--report-trx`) and shows up per failing test in the GitHub test report. Alternatively both `Playwright.Axe` and `Deque.AxeCore.Playwright` ship an HTML report option (`AxeHtmlReportOptions(reportDir)` / `AxeHtmlReportOptions`). Either one deletes the statics, the lock and the fixture change in one move.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -29,4 +71,74 @@ protected async Task NavigateToTestAppAsync(string relativePath)
...
+				"wcag2aa",
+				"wcag21a",
+				"wcag21aa",
+				"EN-301-549"
+			}),
+
+			// Specify rules.
+			rules: new Dictionary<string, AxeRuleObjectValue>()
+			{
+				// ignore problematic rules, currently none applicable
```
hakenr, line 90: The comment says "currently none applicable" but the very next line disables `document-title` (WCAG 2.4.2, Level A). The real cause is in TestApp: `Havit.Blazor.TestApp/Components/App.razor` has `<HeadOutlet />` but no `<title>` (only `Error.razor` sets a `PageTitle`). Please add `<title>Havit.Blazor.TestApp</title>` to `App.razor` and drop this override, rather than suppressing the rule for every page. If a suppression really is needed later, record the reason next to it, otherwise the next person deletes the "empty" rules block and the suite turns red.

Havit.Blazor.E2ETests/TestAppTestBase.cs
```diff
@@ -29,4 +71,74 @@ protected async Task NavigateToTestAppAsync(string relativePath)
...
+			{
+				var description = violation.Description?.Replace("\n", " ").Replace("|", "\\|");
+				var impact = violation.Impact.ToString();
+
+				foreach (var node in violation.Nodes)
+				{
+					var target = string.Join(", ", node.Target ?? new List<string>());
+					target = target.Replace("|", "\\|");
+
+					message.AppendLine($"| {TestContext.Current.Test?.TestDisplayName} | {violation.Id} | {description} | {impact} | {target} |");
```
hakenr, line 130: Nit, latent today: only `description` gets `\n` and `|` escaped and `target` only `|`; the test display name and `violation.Id` are interpolated raw, and `\r` is never handled. The project has only `[Fact]`s now, but the first `[Theory]` whose rendered display name contains `|` or a newline splits the row into extra columns and corrupts both the report and the assertion message. One small `EscapeCell` helper (strip `\r`, `\n` to space, `|` to `\|`) applied to all five columns fixes it. Moot if you move to attachments / the library report.

Havit.Blazor.E2ETests/TestAppAssemblyInitializer.cs
```diff
@@ -36,5 +37,28 @@ public TestAppAssemblyInitializer()
...
 		_factory?.Dispose();
+
+		lock (TestAppTestBase.Lock)
+		{
+			var filePath = Path.Combine(
+				Path.GetTempPath(),
+				$"axe-results-{DateTime.UtcNow:yyyyMMdd-HHmmss}.md"
+			);
+
+			if (TestAppTestBase.AxeReport.Length == 0)
```
hakenr, line 48: This block runs on every test run, whether or not axe was compiled in. I ran `HxAlertTests` in plain Debug: zero axe checks executed, and a 15-byte `axe-results-*.md` containing `No Errors Found` still appeared in `%TEMP%`. That is a false-positive artefact (nothing was checked), it contradicts CONTRIBUTING.md ("written to the temporary folder if violations are found"), and every local and CI run leaves another timestamped file behind. Also the "Axe report written to" line is only printed in the violations branch, so the file is effectively invisible.

At minimum gate this on the same toggle as `RunAxe` and do not write a file when there is nothing to report. Better: drop the fixture change altogether in favour of per-test attachments (see my comment on `AxeReport`).

Havit.Blazor.E2ETests/TestAppAssemblyInitializer.cs
```diff
@@ -36,5 +37,28 @@ public TestAppAssemblyInitializer()
 	public void Dispose()
 	{
 		_factory?.Dispose();
```
hakenr, line 39: If this throws (host `StopAsync` timeout -> `OperationCanceledException`, or an `ObjectDisposedException` on a double dispose), the whole report block below is skipped and the summary file for a run that did find violations is silently lost, exactly in the flaky-teardown case. If the report stays here, wrap this in `try { _factory?.Dispose(); } finally { /* write report */ }`.

hakenr (changes requested):
Thanks for pushing this forward, and for the honest PoC framing. I checked the branch out and ran it: both configurations build with zero warnings, and under `-c ACCESSIBILITYTESTS` axe runs after each test (4 of 7 `HxModal_Basic_Tests` fail on real `color-contrast` findings, report lands in `%TEMP%`, no orphaned headless Chrome left behind on my machine). The mechanism works.

Before this can merge, the shape needs to change in a few places. The big ones, details inline:

1. **Docs and code disagree on the switch.** CONTRIBUTING.md says the `ACCESSIBILITYTESTS` environment variable enables axe; the code only has a compile-time `#if` tied to a new build configuration. Following the docs runs zero checks and still writes a "No Errors Found" report.
2. **Please drop the solution-wide build configuration** and read an environment variable at runtime instead. That removes the slnx/csproj/`#if` changes, avoids building all 38 projects into a third `bin/ACCESSIBILITYTESTS/` flavour, lets CI reuse its Release binaries, and makes the docs true.
3. **Switch to `Deque.AxeCore.Playwright`.** `Playwright.Axe` has not had a release since July 2022 and bundles axe-core 4.4.3, which does not know the `EN-301-549` tag you configure (it silently matches nothing). Deque's package is maintained (4.13, axe-core 4.13) with a near-identical API.
4. **Consider a dedicated route-based `[Theory]` instead of hooking `DisposeAsync`.** Today 118 tests scan 37 of 86 TestApp routes (some 9x) and 49 routes are never audited; a per-route theory covers each page once and lets us `Skip` pages while burning findings down. Happy to discuss this one, it is the design question you raised.

Smaller items inline: report written on every run, axe running after already-failed tests, exception ordering in the two Dispose methods, the `document-title` suppression, and Markdown escaping.

The `core.autocrlf` tip is accurate for Windows contributors (there are real Roslyn IDE0055 line-ending bugs), but it is unrelated to this PR; please move it to its own PR.

PR #1414, opened 2026-05-29
[HxGrid] Fall back to default sorting when CurrentUserState has no sorting (#1413)

### LINKED ISSUE #1413: [HxGrid] Default sort column is lost after CurrentUserState reset — DataProvider receives empty Sorting
cc @jirikanda 

## Describe the bug

`HxGrid` applies the default sorting (columns marked `IsDefaultSortColumn`) **only on the first render** (`OnAfterRenderAsync(firstRender: true)`).

After the first render, whenever the current sorting is re-resolved to "none" — e.g. the bound `CurrentUserState` is reset to a state with `Sorting == null`, or `RefreshDataAsync(GridStateResetOptions.ResetSorting)` is invoked at a moment the columns are not available — the grid does **not** fall back to the default sort column again. It then issues a `DataProvider` request with an **empty `Sorting`** even though a default sort column is defined.

This is inconsistent: a grid that declares a default sort column is expected to sort by it whenever no explicit sorting is set, not only on the very first load.

### Where it comes from

In `HxGrid.OnParametersSetAsync`:

```csharp
if (_previousUserState != CurrentUserState)
{
    _currentSorting = DeserializeCurrentUserStateSorting(CurrentUserState.Sorting); // -> null when Sorting is null/empty
}
...
if (shouldRefreshData)
{
    await RefreshDataCoreAsync(); // request built with empty _currentSorting
}
```

`DeserializeCurrentUserStateSorting(null)` returns `null`, and the default-sorting fallback exists only in the `firstRender` branch of `OnAfterRenderAsync`, so it is never re-applied.

## Expected behavior

When no explicit sorting is set and the grid defines a default sort column (`IsDefaultSortColumn`), the grid should sort by the default column for **every** data load, so the `DataProvider` request never contains empty sorting while a default sort column exists.

## Impact

Consumers that rely on a non-empty sorting (e.g. server-side queries that require a deterministic `ORDER BY`, or that contract-check the sorting) receive an empty/`null` sorting and fail. With protobuf-net based gRPC data providers an empty `SortItem[]` is additionally indistinguishable from `null` on the server.

## Steps To Reproduce

1. Render an `HxGrid` with `@bind-CurrentUserState`, `PageSize`, and a column with `IsDefaultSortColumn="true"` + `SortString`.
2. Let it load (DataProvider receives the default sorting — correct).
3. Reset the bound state: `CurrentUserState = new GridUserState();` (i.e. `Sorting = null`).
4. The grid reloads and the `DataProvider` now receives an **empty** `Sorting` instead of the default sort column.

A deterministic E2E repro page + test is added in the accompanying PR.

## Version

4.24.3 (reproduces on current `main`).
@

hakenr:
Fixes #1413.

## Problem

`HxGrid` applies the default sorting (columns marked `IsDefaultSortColumn`) **only on the first render** (`OnAfterRenderAsync(firstRender: true)`). After the first render, whenever the current sorting is re-resolved to "none" — e.g. the bound `CurrentUserState` is reset to a state with `Sorting == null` — the grid does **not** fall back to the default sort column again and issues a `DataProvider` request with **empty `Sorting`**, even though a default sort column is defined.

This breaks consumers that rely on a non-empty sorting (e.g. server-side queries requiring a deterministic `ORDER BY`, or that contract-check the sorting). With protobuf-net based gRPC data providers an empty `SortItem[]` is additionally indistinguishable from `null` on the server.

## Repro (first commit)

A TestApp page + Playwright E2E test reproduce it deterministically by recording the `SortString` the `DataProvider` receives on each load:

1. Initial load → `Phone` (the default sort column) ✔
2. Reset `CurrentUserState` to `new GridUserState()` → **`(empty)`** ✘ (the bug)

The committed test asserts the sorting log is `"Phone | Phone"`; before the fix the second (post-reset) request is `"(empty)"`.

Note: `RefreshDataAsync(GridStateResetOptions.ResetSorting)` / `All` already re-derive the default correctly (they call `GetDefaultSorting()`); only the `CurrentUserState`/`OnParametersSetAsync` path was affected.

## Fix (second commit)

Introduce `EnsureCurrentSortingInitializedWithDefault()` and call it at the start of `RefreshDataCoreAsync()` (the single choke point before every data load). When `_currentSorting` is `null` and the deserialization was not postponed (columns registered), it falls back to the default sorting if a default sort column exists. When the grid has **no** default sort column the sorting is intentionally left empty (no behavior change), and an existing/explicit sorting is never overridden.

Column-header clicks can never produce empty sorting (`ApplyColumnToSorting` always returns a non-empty result), so this fallback only affects programmatic "no sorting" states.

## Verification

- New E2E test fails without the fix (`Phone | (empty)`), passes with it (`Phone | Phone`).
- All 96 existing grid unit tests pass (net8.0/9.0/10.0).


hakenr:
@copilot Update from main and migrate the new test(s) to xUnit (main migrated from MSTest to xUnit in meantime).

jirikanda (commented):
Should or should not the EnsureCurrentSortingInitializedWithDefault invoke CurrentUserStateChanged event callback? Like SetCurrentSortingWithEventCallback does. When OnAfterRenderAsync sets sorting it also invokes the callback via SetCurrentSortingWithEventCallback.

hakenr:
Replaced with
* #1425

The real issue was that the consuming project was trying call RefreshDataAsync on already disposed HxGrid.
(GoranG3 / ContactDetail / the HxTabPanel with RenderMode.ActiveTabOnly and the LoadContactDetails method was trying to RefreshDataAsync on HxGrid instance, which was no longer visible - the tab changed.)

PR #1794, opened 2026-09-14
[Havit.Blazor.Storage] Drop the Havit.Core dependency

jirikanda:
`Havit.Blazor.Storage` referenced the whole `Havit.Core` package for a single purpose: argument guards in `BrowserStorageAccessor`. Since .NET 8 the BCL provides equivalent throw helpers, so the dependency no longer pays for itself.

The cost was not just an extra package in the graph. `Havit.Core` is not currently marked as trimmable, so anything referencing `Havit.Blazor.Storage` pulls its full transitive closure into the output — `System.Net.Mail` among others — with no way for the linker to remove it. That is a real penalty for a Blazor WebAssembly app, where the payload is downloaded to the browser and this library is primarily meant to be used. `Havit.Core` is expected to become trimmable, but there is no reason for a storage-access library to wait on that when the only thing it needed were null checks.

## Changes

- Replaced all 13 `Contract.*` calls in `BrowserStorageAccessor` with BCL throw helpers:
  - 9× `Contract.Requires<ArgumentException>(!string.IsNullOrWhiteSpace(key))` → `ArgumentException.ThrowIfNullOrWhiteSpace(key)`
  - 4× `Contract.Requires<ArgumentNullException>(value is not null)` → `ArgumentNullException.ThrowIfNull(value)`
- Removed `<PackageReference Include="Havit.Core" />` from `Havit.Blazor.Storage.csproj` and the corresponding `using Havit.Diagnostics.Contracts;`.
- Bumped `StoragePackagesVersion` to 1.0.2.

`BrowserStorageAccessor` was the only file in the project using `Havit.Core`, so this removes the dependency entirely. No other project is affected.

## Behavioral notes

**Exception type for a `null` key.** `Contract.Requires<ArgumentException>` threw `ArgumentException` for both `null` and whitespace. `ArgumentException.ThrowIfNullOrWhiteSpace` distinguishes the two and throws `ArgumentNullException` for `null`. Since `ArgumentNullException` derives from `ArgumentException`, existing `catch (ArgumentException)` blocks are unaffected; only an exact-type check would notice. There are no tests asserting on these exceptions (`Havit.Blazor.Storage.E2ETests` drives the TestApp and does not cover argument validation).

Related upside: the new helpers populate `ParamName` via `CallerArgumentExpression`, which `Contract.Requires` did not — exception messages are now more specific.

**Boxing in the generic overloads.** In `SetValue<TValue>` and `SetValueAsync<TValue>`, `value` is an unconstrained `TValue` while `ArgumentNullException.ThrowIfNull` takes `object`, so value types get boxed on every call. Semantics are unchanged (a boxed struct is never null, just as `value is not null` never threw for one), but it is an extra allocation. I kept the guards 1:1 with the original code; if we would rather avoid it, both generic overloads immediately delegate to their `string`-typed counterparts, where the check is repeated anyway — so they could simply be dropped. Happy to do that if reviewers prefer.

## Verification

`dotnet build` succeeds for all three target frameworks (net8.0, net9.0, net10.0) with 0 warnings and 0 errors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


hakenr:
@jirikanda I'm not getting the point. How many projects do we have (or how many do we expect around the world) where `Havit.Blazor.Storage` will be used and `Havit.Core` not?

On the other hand - do you expect these "new" helpers will be our new future coding standard (default)?

jirikanda:
As an open-source / hobby developer using `Havit.Blazor.Storage`, I was surprised
to see my Blazor app downloading `System.Net.Mail.<hash>.wasm`.
The chain is: `Havit.Blazor.Storage` → `Havit.Core` → shared framework → `System.Net.Mail`.

`Havit.Blazor.Storage` is not marked as trimmable (but it can and should be).
Until then you can opt it into trimming from your own csproj:

```xml
<ItemGroup>
  <TrimmableAssembly Include="Havit.Blazor.Storage" />
</ItemGroup>
```

`Havit.Core` is not marked as trimmable either. That will change soon (today) —
you can't get rid of `Havit.Core` the reference from all packages anyway, it's part of the public API of `HxGrid` etc.

> On the other hand - do you expect these "new" helpers will be our new future coding standard (default)?

Absolutely 💪

